PR #21, opened 2019-01-11
Support for GEL theming in the docz app

jossmac:
- delete media-queries package
- express breakpoints as brand token
- move facepaint logic to grid package
- move applicable deps to `peerDependencies`, e.g. "prop-types", "react", "react-dom"
- brands only have default export object
- add docs for tabcordion
- improve docs for icon


components/core/package.json
```diff
@@ -19,6 +19,7 @@
 		"@emotion/core": "^10.0.4",
 		"@emotion/styled": "^10.0.4",
 		"emotion-theming": "^10.0.4",
+		"prop-types": "^15.6.2",
```
dominikwilkowski: should also be peer

components/media-queries/package.json
```diff
@@ -16,7 +16,8 @@
 	},
 	"main": "dist/index.js",
 	"dependencies": {
-		"facepaint": "^1.2.1"
+		"facepaint": "^1.2.1",
+		"prop-types": "^15.6.2"
```
dominikwilkowski: should be peer

package.json
```diff
@@ -71,6 +71,7 @@
 		"emotion-theming": "^10.0.4",
 		"facepaint": "^1.2.1",
 		"prettier": "^1.15.3",
+		"prop-types": "^15.6.2",
```
dominikwilkowski: peer?

components/media-queries/package.json
jossmac: will delete this package monday

components/core/package.json
jossmac: fixed

components/media-queries/package.json
dominikwilkowski: @jossmac deleted?

jossmac: yes

PR #27, opened 2019-01-16
Remove emotion-theming

jossmac:
Replaced by a bespoke context implementation


components/core/src/Theme.js
```diff
@@ -1,10 +1,61 @@
...
-export const GEL = ({ brand, ...props }) => <ThemeProvider theme={brand} {...props} />;
+// TODO: is there a useful "default" value we should add here?
+export const ThemeContext = createContext();
+
+export const useTheme = () => {
+	const themeObj = useContext(ThemeContext);
+
+	if (!themeObj) {
+		throw new Error(
+			`It looks like you're trying to use a GEL component without wrapping your application with the theme provider.`
```
dominikwilkowski: `wrapping your application with the theme provider` mh 🤔  we export the theme provider as `GEL` so we should probably call it that in the error message now?

jossmac: Replaced with:

> "GEL components require that you wrap your application with the \<GEL \/\> theme provider from @westpac/core."

PR #38, opened 2019-07-09
WIP:  Adding Button, ButtonGroup, Badge, Well, Table components

jonnystening:
Also includes some other environment tweaks...

### Tokens
• Token files have a Color dependency (may be promoted to a dependency of Core rather than at a token level?)
• Token files have new token key/values; used by these new components

### Helpers
• `helpers/example/index.js`'s styled component `Container` commented out; `Container` now imported from `Grid` so we're using a consistent grid layout to display these component examples
• `helpers/webpack.config.js`'s function `labelFromSlug` commented out, edited to return sentence case (rather than title case); used to power the example testing environment sidebar nav
• `helpers/.component-template/src/__COMPONENT_NAME__.js` updated to include theme \

### Icon component
• `Icon` component wrapper name updated (`Wrapper` --> `IconWrapper`); props now spread


dominikwilkowski:
Let's fix tests before merging :)

dominikwilkowski:
as in: just run `yarn format` please ;)

jonnystening:
haha thanks @dominikwilkowski

dominikwilkowski:
If this is still WIP then put `WIP: ` in the title of the PR and the WIP bot will mark it as no-mergable so no one accidentally merges it

components/button/src/Button.js
```diff
@@ -0,0 +1,304 @@
...
+	tag: Tag,
+	onClick,
+	...props
+}) => {
+	const theme = useTheme();
+	const mq = mediaqueries(theme.breakpoints);
+
+	// Common styling
+	const styleCommon = {
+		textDecoration: 'none',
```
jaortiz: Refactor styling as discussed (I've sent you my variations file on slack) :nail_care:

components/button/examples/50-active-state.js
```diff
@@ -0,0 +1,78 @@
...
+			</Button>
+		</p>
+
+		<hr />
+
+		<h3>
+			Active buttons with an <code>&lt;a&gt;</code> tag
+		</h3>
+		<p>
+			<Button href="#0" appearance="primary" className="active">
```
jaortiz: Is there ever a use case to be able to force a an active state on a button? I don't see the value of it and not a fan of targeting a specific class on the component when using emotion :nail_care:

components/button/examples/60-disabled-state.js
```diff
@@ -0,0 +1,79 @@
...
+			</Button>
+		</p>
+
+		<hr />
+
+		<h3>
+			Disabled buttons with an <code>&lt;a&gt;</code> tag
+		</h3>
+		<p>
+			<Button href="#0" appearance="primary" className="disabled">
```
jaortiz: For further discussion on whether a button as an 'a' tag can be disabled and if so we should pass disabled as a prop and handle via emotion rather than passing and targeting a className

components/button/src/utils.js
```diff
@@ -0,0 +1,8 @@
+import weakMemo from '@emotion/weak-memoize';
+
+const minWidth = width => `@media (min-width: ${width}px)`;
+const mapBreakpoints = ([key, value]) => minWidth(value);
+
+export const mediaqueries = weakMemo(breakpoints =>
```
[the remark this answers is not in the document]
jonnystening: sure

components/table/src/utils.js
```diff
@@ -0,0 +1,8 @@
+import weakMemo from '@emotion/weak-memoize';
+
+const minWidth = width => `@media (min-width: ${width}px)`;
+const mapBreakpoints = ([key, value]) => minWidth(value);
+
+export const mediaqueries = weakMemo(breakpoints =>
+	Object.entries(breakpoints).map(mapBreakpoints)
+);
```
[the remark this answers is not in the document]
jonnystening: sounds good to me

components/button/examples/50-active-state.js
jonnystening: Possibly when a button is used within a dropdown component? 

It's also provided so we can easily illustrate the active styling in the dev/example environment, and run Cypress integration testing over it

components/button/examples/60-disabled-state.js
jonnystening: I think we should run with a `<a>` (link) cant be disabled in HTML, it's bad UX to do this too.

We have to be careful doing everything 'the React way' because we have to also maintain support for non-react environments via this new Legacy support layer (we're going to be adding down the track). I'd like to leverage default html attributes where available.

components/table/src/utils.js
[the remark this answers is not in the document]
jonnystening: This should be exposed by @westpac/core

components/button/src/Button.js
```diff
@@ -0,0 +1,276 @@
+/* @jsx jsx */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useTheme } from '@westpac/core';
+import { mediaQueries } from './utils';
```
jaortiz: Ive had more thought about this, since this is so similar to the one in grid and we plan to add that to @westpac/core we should only have one and just use that version to avoid confusion and be consistent.

I've played around with it and we can refactor the responsive styling options to use it and it ends up being more compact :grinning:

components/button/src/Button.js
```diff
@@ -0,0 +1,276 @@
...
+
+	// Block styling
+	const styleBlock = Array.isArray(block)
+		? // Block provided as an array
+		  block.map((b, i) => {
+				return i === 0 ? getBlockStyling(b) : { [mq[i]]: getBlockStyling(b) };
+		  })
+		: // Block prop provided a string, returned as single index array
+		  [getBlockStyling(block)];
+
```
jaortiz: Refactoring to use facepaint we can replace styleSize, getSizeStyling(), styleBlock, getBlockStyling() to
```
// new util
const asArray = val => (Array.isArray(val) ? val : [val]);

const styleResponsive = () => {
		const sizeArr = asArray(size);
		const blockArr = asArray(block);

		const padding = [],
			fontSize = [],
			height = [];

		sizeArr.forEach(s => {
			const pad = button.size[s].padding;
			if (trim) pad[1] = 0;
			padding.push(pad.join(' '));
			fontSize.push(button.size[s].fontSize);
			height.push(button.size[s].height);
		});

		return {
			padding,
			fontSize,
			height,
			display: blockArr.map(b => (b ? 'flex' : 'inline-flex')),
			width: blockArr.map(b => (b ? '100%' : 'auto')),
		};
	};
```

components/button/src/Button.js
```diff
@@ -0,0 +1,276 @@
...
+	icon: Icon,
+	iconPosition,
+	justify,
+	children,
+	tag: Tag,
+	onClick,
+	...props
+}) => {
+	const theme = useTheme();
+	const mq = mediaQueries(theme.breakpoints);
```
jaortiz: replace with common paint util

components/button/src/Button.js
```diff
@@ -0,0 +1,276 @@
...
+	const button = theme.button;
```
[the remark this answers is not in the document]
jonnystening: NIce!

Don't even need `theme` in this example... 

```
const { button, breakpoints } = useTheme();
const mq = mediaQueries(breakpoints);
````

File: components/button/src/Button.js (hunk 6 above)
Comment (jonnystening) [reply components/button/src/Button.js]: Great, keen to see it

File: components/button/src/Button.js (hunk 8 above)
Comment (jonnystening) [reply components/button/src/Button.js]: Will we always need to `@emotion/weak-memoize` or only for grid?

File: components/button/src/Button.js [hunk 10]
```diff
@@ -0,0 +1,310 @@
...
+	const styleBlock = Array.isArray(block)
+		? // Block provided as an array
+		  block.map((b, i) => {
+				let bp = mq[i];
+				return i === 0 ? getBlockStyling(b) : { [bp]: getBlockStyling(b) };
+		  })
+		: // Block prop provided a string, returned as single index array
+		  [getBlockStyling(block)];
+
+	if (props.href && Tag === 'button') {
```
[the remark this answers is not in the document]
jonnystening: We agreed to make it simply

```
if (props.href) 
```

PR #74, opened 2019-07-11
Feature/alert

florettewest:


components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+						{Icon && <Icon size="small" className="icon icon-alert" />}
+						<div className="alert-body">{children}</div>
+						{Icon && (
+							<button
+								onClick={() => {
+									setAnim(false);
+								}}
+							>
+								<CloseIcon size="small" />
+							</button>
```
jonnystening: I think this should use our Button component with appearance="link" and icon={CloseIcon}, like how we do it in GUI2.0 where the alert box has a dependency on the button component.

```
<Button appearance="link" icon={CloseIcon} onClick={() => { setAnim(false); }} />
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+					in={animAlert}
+					classNames="anim"
+					timeout={400}
+					onExited={() => setOpen(false)}
+				>
+					<div
+						css={[styleAppearance(appearance), styleBox(appearance), styleOverrides]}
+						className={appearance === 'system' ? 'system' : ''}
+					>
+						{Icon && <Icon size="small" className="icon icon-alert" />}
```
jonnystening: This should use responsive sizing like how's it's done in GUI2.0

```
{Icon && <Icon size={['small', 'medium']} />}
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+
+		a: {
+			color: 'inherit',
+		},
+	});
+};
+
+const styleBox = appearance => {
+	const theme = useTheme();
+	return css({
```
jonnystening: Lets not use Emotion's css function like this; return styling in simple object notation

```
return {
   ...
};
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+		position: 'relative',
+		marginBottom: '21px',
+		zIndex: '1',
+		padding: '18px 34px 18px 18px',
+		borderTop: `1px solid ${Color('white')
+			.mix(Color(theme.colors[appearance]), 0.5)
+			.hex()}`,
+		borderBottom: `1px solid ${Color('white')
+			.mix(Color(theme.colors[appearance]), 0.5)
+			.hex()}`,
```
jonnystening: much of these 'design' settings need to be set in the tokens rather than stored in the Alert component

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+			border: 'none',
+			position: 'absolute',
+			right: '0',
+			top: '15px',
+			padding: '5px 12px',
+
+			'&:hover': {
+				cursor: 'pointer',
+				opacity: '0.5',
+			},
```
jonnystening: much of this styling would be unnecessary if the button was powered by our Button component itself, and the Button positioning should really be provided inline, rather than via a button child selector

components/alert/package.json
```diff
@@ -0,0 +1,41 @@
+{
+	"name": "@westpac/alert",
+	"version": "0.0.1",
+	"description": "Grid component for the Westpac GEL",
```
jonnystening: please update this to something alert-esque

components/alert/package.json
```diff
@@ -0,0 +1,41 @@
...
+		{
+			"name": "Dominik Wilkowski",
+			"email": "Hi@Dominik-Wilkowski.com",
+			"url": "https://dominik-wilkowski.com"
+		},
+		{
+			"name": "Joss Mackison",
+			"email": "joss@thinkmill.com.au",
+			"url": "http://jossmackison.me"
+		}
```
jonnystening: where is Flore?

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
+/* @jsx jsx */
+
+import React, { useState } from 'react';
+import { css, jsx, useTheme } from '@westpac/core';
+import Color from 'color';
```
jonnystening: I don't think this will required; see comment re moving styling to tokens

components/alert/README.md
```diff
@@ -0,0 +1,3 @@
+# @westpac/icon
+
+The icon component.
```
jonnystening: update for alert

components/alert/docs.mdx
```diff
@@ -0,0 +1,132 @@
...
+				<Glyph />
+				<Name>{name}</Name>
+			</Cell>
+		))}
+	</Grid>
+</Playground>
+
+## Props
+
+<PropsTable of={Icon} />
```
jonnystening: update for alert

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+
+import { CloseIcon } from '../../icon/src'; //until icon package is published
```
jonnystening: wont be required; see comment re using our Button component

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+// ==============================
+// Utils
+// ==============================
+
+// ==============================
+// Component
+// ==============================
+const styleAppearance = appearance => {
+	const theme = useTheme();
+	return css({
```
jonnystening: Lets not use Emotion's css function like this; return styling in simple object notation

```
return {
   ...
};
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+							</button>
+						)}
+					</div>
+				</CSSTransition>
+			)}
+		</>
+	);
+};
+
+const AlertText = ({ appearance, icon: Icon, iconSize, children, as: As }) => {
```
jonnystening: use tag as the prop to set the tag type

```
const AlertText = ({ appearance, icon: Icon, iconSize, children, tag: Tag }) => {
```

NB. This AlertText may not be required; will discuss with Justin

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+	'&.system': {
+		backgroundColor: 'yellow',
+		borderColor: 'yellow',
+		color: 'black',
+	},
+});
+
+const AlertBox = ({ appearance, icon: Icon, children }) => {
+	const [open, setOpen] = useState(true);
+	const [animAlert, setAnim] = useState(true);
```
jaortiz: this animation state variable isnt required, see below

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+
+	return (
+		<>
+			{open && (
```
jaortiz: The conditional open && isnt required. The library can handle the component mounting and unmounting for us. Replace with
```
<CSSTransition in={open} unmountOnExit classNames="anim" timeout={400}>
{ code }
</CSSTransition>
```
Here we use the state of whether the alert is visible or not to handle the mounting and the animation

components/alert/src/Alert.js
```diff
@@ -0,0 +1,146 @@
...
+					<div
+						css={[styleAppearance(appearance), styleBox(appearance), styleOverrides]}
+						className={appearance === 'system' ? 'system' : ''}
+					>
+						{Icon && <Icon size="small" className="icon icon-alert" />}
+						<div className="alert-body">{children}</div>
+						{Icon && (
+							<button
+								onClick={() => {
+									setAnim(false);
```
jaortiz: change to setOpen(false)

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+		position: 'relative',
+		flex: 1,
+		top: ['auto', 2],
+	};
+
+	return (
+		<CSSTransition in={open} unmountOnExit classNames="anim" timeout={400}>
+			<div css={mq({ ...styleCommon, ...styleAppearance })}>
+				{closable && (
+					<Button
```
jaortiz: I dont think we should be using the GEL button for this. This should be a regular html button as this is a very simple element. Using the GEL button introduces a new unncecessary dependency and all the extra code that comes with it.

jonnystening: @jaortiz Using Button ensures everything button related is considered, and managed via the one source-of-truth component. If we roll our own for Alert we have to make sure it does everything Button does incl. it's handling of icons and focus styling etc.

This is also the case with Modal. Keen to discuss...

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+
+export const Alert = ({ appearance, closable, icon: Icon, children }) => {
+	const [open, setOpen] = useState(true);
+	const { breakpoints, alert } = useTheme();
+	const mq = paint(breakpoints);
+
+	const styleCommon = {
+		marginBottom: alert.marginBottom,
+		padding: alert.padding,
+		paddingRight: [18, 30],
```
jonnystening: Should use tokens

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+		position: 'relative',
+		display: ['block', 'flex'],
+		zIndex: 1,
+		transition: 'opacity 0.3s ease-in-out',
+
+		a: {
```
jonnystening: Also style for headings etc...

```
'a, h1, h2, h3, h4, h5, h6, ol, ul': {
	color: 'inherit',
},
```

brands/BOM/tokens/index.js
```diff
@@ -545,4 +546,66 @@ export default {
...
 	},
+
+	alert: {
+		padding: '18px',
+		marginBottom: '21px',
+
+		appearance: {
+			success: {
+				color: COLORS.success,
+				backgroundColor: Color('white')
```
jonnystening: Use our new tint() function from utils.js

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+			color: 'inherit',
+		},
+
+		'&.anim-exit-active': {
+			opacity: 0,
+		},
+	};
+
+	const styleAppearance = {
+		...alert.appearance[appearance],
```
jonnystening: I would simplify the alert tokens to just color, backgroundColor and borderColor. Then set up the components appearance styling as...
```
const styleAppearance = {
	color: alert.appearance[appearance].color,
	backgroundColor: alert.appearance[appearance].backgroundColor,
	borderTop: `${alert.borderWidth} solid`,
	borderBottom: `${alert.borderWidth} solid`,
	borderColor: alert.appearance[appearance].borderColor,
};
```

with token shape...
```
success: {
	color: COLORS.success,
	backgroundColor: tint(COLORS.success, 0.1),
	borderColor: tint(COLORS.success, 0.5),
},
```

The alert borders are always going to be the same border width top and bottom, and always solid border style. The tokens don't need this granularity.

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+	};
+
+	const styleClose = {
+		color: alert.appearance[appearance].color,
```
jonnystening: Close button can make use of `inherit` color value
```
const styleClose = {
	color: 'inherit',
	...
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+	};
+
+	const styleIcon = {
+		float: ['left', 'none'],
+		marginRight: [6, 12],
+		flex: 'none',
+		fontSize: [18, 24],
+		width: [18, 24],
+		height: [18, 24],
+	};
```
jonnystening: The Button component should set all these icon sizing settings. We shouldn't promote ad-hoc button and icon styling, these should be managed in the one place.

```
const styleIcon = {
	float: ['left', 'none'],
	marginRight: [6, 12],
	flex: 'none',
};
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+
+	const styleBox = {
```
jonnystening: `styleBody` probably more suitable

```
const styleBody = {
	...
}
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+		position: 'relative',
+		flex: 1,
+		top: ['auto', 2],
```
jonnystening: The `top` tweak is only added if there is an icon present. It helps the text align vertically against the icon.

```
const styleBody = {
	position: 'relative',
	flex: 1,
	top: [null, Icon ? 2 : null],
};
```

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+						onClick={() => {
+							setOpen(false);
+						}}
+						icon={CloseIcon}
+						size="small"
+						appearance="link"
+						css={styleClose}
+					/>
+				)}
+				{Icon && <Icon className="icon icon-alert" css={mq({ ...styleIcon })} />}
```
jonnystening: remove `className`, it isn't required

components/alert/src/Alert.js
```diff
@@ -0,0 +1,128 @@
...
+// Types
+// ==============================
+
+Alert.propTypes = {
+	/**
+	 * The alert appearance.
+	 *
+	 * Defaults to "information"
+	 */
+	appearance: PropTypes.oneOf(['success', 'information', 'warning', 'danger', 'system']),
```
jonnystening: I've been pulling out any `PropTypes.oneOf` options into an options const. I think it's nice pattern to have all of these in the one spot. Thoughts?

```
const options = {
	appearance: ['success', 'information', 'warning', 'danger', 'system'],
};

Alert.propTypes = {
	/**
	 * The alert appearance.
	 *
	 * Defaults to "information"
	 */
	appearance: PropTypes.oneOf(options.appearance),
	...
```

components/alert/src/Alert.js
florettewest: We agreed to leave it as is...

components/alert/src/Alert.js
florettewest: We decided to leave it as is...

components/alert/src/Alert.js
florettewest: This code actually applies to the icon before the text...

components/alert/src/Alert.js
florettewest: Hey Jonny, my code is now looking like: `padding: getPaddingSM(),`
If I compare to yours, you have: `[alert.padding, getPaddingSM()]`
The added padding if closeable only applies to sm res and up?

jonnystening: ```padding: [alert.padding, getPaddingSM()]``` ensures for XS we use the default 18px padding, but for SM+ if there is a close button it will use 30px on the right; otherwise it will simply use the default 18px.

What you have will apply the 30px padding for all breakpoints if there's a close button. In XS we float the close button to the right, so the text wraps around the button. The button is no longer pos: abs, so we dont need additional padding.

components/alert/src/Alert.js
```diff
@@ -17,17 +17,18 @@ export const Alert = ({ appearance, closable, icon: Icon, children }) => {
...
 	const { breakpoints, alert } = useTheme();
 	const mq = paint(breakpoints);
+	const getPaddingSM = () =>
+		closable ? `${alert.padding} 30px ${alert.padding} ${alert.padding}` : alert.padding;
 
 	const styleCommon = {
 		marginBottom: alert.marginBottom,
-		padding: alert.padding,
-		paddingRight: [18, 30],
+		padding: getPaddingSM(),
```
jonnystening: ^ We need to disregard the extra RHS padding in XS; it doesn't matter about space for the button in XS since it's floating. We're better to make use of all the space we have for the text.

PR #77, opened 2019-07-17
Modal

jaortiz:


brands/BOM/tokens/index.js
```diff
@@ -493,4 +493,33 @@ export default {
...
 	},
+
+	modal: {
+		backgroundColor: '#fff',
+		borderRadius: 3,
+		width: {
+			small: 300,
+			medium: 600,
+			large: 900,
+		},
```
jonnystening: What I've been trying to do with tokens is ensure that the last key used in each token setting is always a CSS property. That way it's more consistent, and can be also spread. I reckon this should be like this, which is more consistent with other component's tokens with size (ie. button)...

```
modal: {
	size: {
		small: {
			width: 300,
		},
		medium: {
			width: 600,
		},
		large: {
			width: 900,
		},
	},
	...
},
```

components/modal/examples/10-sizing.js
```diff
@@ -0,0 +1,90 @@
+import React, { useReducer } from 'react';
```
jonnystening: Can we use 'size' as the example's menu item... as per buttons etc

`examples/10-size.js`

components/modal/examples/00-default.js
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react';
+import { Modal, ModalHeader, ModalBody, ModalFooter } from '../src';
+
+// TO DO: Replace with proper button import once published to npm
+import { Button } from '../../button/src';
```
jonnystening: Should we write this TODO comment consistently... I've got many of these to switch over once published. Find/replace would be easier?

components/modal/package.json
```diff
@@ -0,0 +1,40 @@
...
+		"test:test-dev": "cypress open"
+	},
+	"main": "dist/index.js",
+	"dependencies": {
+		"@westpac/core": "^0.0.1",
+		"@westpac/icon": "^0.0.1",
+		"prop-types": "^15.6.2",
+		"react-focus-lock": "^2.0.3",
+		"react-transition-group": "^4.2.1",
+		"shortid": "^2.2.14"
```
jonnystening: missing button dependency?

components/modal/src/Modal.js
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			if (child.type === ModalBody) {
+				return cloneElement(child, {
+					id: bodyId,
+				});
+			}
+		}
+		return child;
+	});
```
jonnystening: Can/should this be refactored as one return and cloneElement call? 🤔

This is probs easier to follow?

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,48 @@
...
+			}}
+			{...props}
+		>
+			<Title id={id}>{children}</Title>
+			<button
+				css={{ border: 'none', padding: 0, '&:hover svg': { opacity: 0.5 } }}
+				onClick={onClose}
+			>
+				<CloseIcon size="small" color={colors.text} />
+			</button>
```
jonnystening: can we use our `Button`?

components/modal/src/styled.js
```diff
@@ -0,0 +1,80 @@
...
+		maxHeight: '85%',
+		backgroundColor: modal.backgroundColor,
+		borderRadius: modal.borderRadius,
+		boxShadow: '0 5px 15px rgba(0,0,0,0.5)',
+		transition: 'all 0.3s ease',
+		width: [
+			'auto',
+			...(size === 'small' ? [modal.width.small] : [modal.width.medium]),
+			...(size === 'large' ? [modal.width.large] : []),
+		],
```
jonnystening: why conditionally return and spread single index arrays?

jonnystening: If we refactor the token to how I mentioned above can't we just do...

```
const styles = {
	overflow: 'auto',
	maxHeight: '85%',
	backgroundColor: modal.backgroundColor,
	borderRadius: modal.borderRadius,
	boxShadow: '0 5px 15px rgba(0,0,0,0.5)',
	transition: 'all 0.3s ease',

	width: [
		'auto',
		modal.size[size].width,
	],
	...
}
```

components/modal/src/styled.js
```diff
@@ -0,0 +1,80 @@
...
+	};
+	return <div css={styles} {...props} />;
+};
+
+export const StyledModal = ({ size, ...props }) => {
+	const { modal, breakpoints } = useTheme();
+	const mq = paint(breakpoints);
+
+	const styles = {
+		overflow: 'auto',
```
jonnystening: Needs a 12px margin so it doesn't reach the viewport edge on XS. GUI2.0 has a 12px edge gap using left/right values.

components/modal/src/styled.js
jonnystening: Great way to use of facepaint! 👍

components/modal/src/ModalHeader.js
jonnystening: ^ will also help address the button click area. This button is way too small

components/modal/src/styled.js
jonnystening: Also need to be able to scroll modal content. Resize viewport to short height and check with a modal with long content.

components/modal/src/ModalFooter.js
```diff
@@ -0,0 +1,20 @@
...
+	const { modal } = useTheme();
+
+	return (
+		<div
+			css={{
+				backgroundColor: modal.footer.backgroundColor,
+				borderTop: `${modal.footer.borderWidth} solid ${modal.footer.borderColor}`,
+				textAlign: 'right',
+				padding: modal.footer.padding,
+			}}
```
jonnystening: Need to style adjacent buttons in the footer... they shouldn't touch

components/modal/src/styled.js
jonnystening: Would be...

```
 width: [
 	'auto',
	...(size === 'small' ? [modal.size.small.width] : [modal.size.medium.width]),	
	...(size === 'large' ? [modal.size.large.width] : []),
 ],
```
if you change those tokens

components/modal/examples/00-default.js
jaortiz: I think it would be good to do it consistently so its easier to update once we start publishing packages

jaortiz: Actually its not necessary. We can straight up install the dependency via bolt. Bolt creates a symlink to the local package. Pretty cool stuff :astonished:

components/modal/src/Modal.js
jaortiz: Refactoring it to one return in this case would involve more logic and nested ternaries making it harder to read.

components/modal/src/ModalFooter.js
jaortiz: How much distance? GEL 2 has 2px spacing but it seems small...

components/modal/src/ModalHeader.js
jaortiz: Fixed the button sizing. 

I still believe that it is unnecessary to use the GEL button for simple things like this. It would be good to get ThinkMill's opinion on this. If they think we should use the GEL button then I will change to use it otherwise I will leave it as is for now.

components/modal/src/ModalFooter.js
jaortiz: Also do we think this is necessary? Modal Footer is just a wrapper for styling purposes. 
Should we just let the devs implement the button spacing since we dont know how many buttons they will put in the footer?

components/modal/src/Modal.js
jonnystening: Yeah I had a go. Came out longer 😢

components/modal/src/ModalHeader.js
jonnystening: Hmm

components/modal/src/ModalFooter.js
jonnystening: Yeah we need to mandate and provide this styling or it will likely be implemented inconsistently 

Buttons are inline-block and in GUI2.0 we're assuming there is a space character in-between which we don't have to worry with React/jsx 🙌

Should be 6px gap

```
& + & {
	marginLeft: 6,
}
```

components/modal/src/ModalFooter.js
```diff
@@ -0,0 +1,23 @@
...
+
+	return (
+		<div
+			css={{
+				backgroundColor: modal.footer.backgroundColor,
+				borderTop: `${modal.footer.borderWidth} solid ${modal.footer.borderColor}`,
+				textAlign: 'right',
+				padding: modal.footer.padding,
+				'button + button ': {
+					marginLeft: 2,
```
jonnystening: make it `marginLeft: 6`

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,54 @@
...
+	return (
+		<div
+			css={{
+				display: 'flex',
+				justifyContent: 'space-between',
+				alignItems: 'center',
+				overflow: 'hidden',
+				borderBottom: `${modal.header.borderWidth} solid ${modal.header.borderColor}`,
+				padding: modal.header.padding,
+			}}
```
jonnystening: Button needs to sit at a consistent top/right position. Check out the small modal example. The button is moving down.

You could `alignItems: flex-start` it, or `position: absolute` it. Need to make sure that no matter what heading text is in there the button doesnt move, and the text cant show underneath the button.

jonnystening: _see comment below_

Put `alignItems: flex-start` on the flex parent (this ModalHeader wrapping the flex children) so both the title and the button are top aligned.

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,55 @@
...
+			{...props}
+		>
+			<Title id={id}>{children}</Title>
+			<button
+				css={{
+					border: 'none',
+					padding: '5px 12px',
+					cursor: 'pointer',
+					touchAction: 'manipulation',
+					alignSelf: 'flex-start',
```
jonnystening: You can remove this (_see above_). The flex parent can control the title and button children, both should align to the top really.

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,55 @@
...
+// ==============================
+export const ModalHeader = ({ id, onClose, children, ...props }) => {
+	const { modal, colors } = useTheme();
+
+	return (
+		<div
+			css={{
+				display: 'flex',
+				justifyContent: 'space-between',
+				alignItems: 'center',
```
jonnystening: `alignItems: flex-start`

components/modal/src/ModalHeader.js
jonnystening: Spoke to Dom, he said that we should be reusing our components and be "atomic". Build small blocks and reuse them throughout. Any file size and memory usage etc concerns negligible 🤷‍♂️

I think we should implement our Button and mark this as a thing to discuss in depth.

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,57 @@
...
+			>
+				<CloseIcon size="small" color={colors.text} />
+			</button>
+		</div>
+	);
+};
+
+ModalHeader.propTypes = {
+	/** id of the title for accessibility */
+	id: PropTypes.string,
```
jonnystening: .isRequired?

components/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,57 @@
...
+	/** onClose function for the close button */
+	onClose: PropTypes.func,
+};
+
+ModalHeader.defaultProps = {
+	id: null,
```
jonnystening: Do we need to do this? If so should do throughout 🤔

components/modal/src/styled.js
```diff
@@ -0,0 +1,80 @@
...
+		},
+
+		'&.modal-appear-done': {
+			transform: 'translate(0px,30px)',
+		},
+	};
+
+	return <div css={mq(styles)} {...props} />;
+};
+
```
jonnystening: We should apparently checking proptypes for all components, even these simple styled components.

components/modal/src/ModalFooter.js
```diff
@@ -0,0 +1,23 @@
...
+	const { modal } = useTheme();
+
+	return (
+		<div
+			css={{
+				backgroundColor: modal.footer.backgroundColor,
+				borderTop: `${modal.footer.borderWidth} solid ${modal.footer.borderColor}`,
+				textAlign: 'right',
+				padding: modal.footer.padding,
+				'button + button ': {
```
jonnystening: `'button + button': {`

extra space char 🤷‍♂️

components/modal/src/ModalHeader.js
jaortiz: I already handle the aria-describedby and aria-labelledby by auto-generating an id an passing to appropriate modal components

components/modal/src/styled.js
jaortiz: I dont see the point as I am composing these are not exported in the package. They are only consumed in other components which I am composing.

components/modal/src/ModalHeader.js
jaortiz: We need to introduce a prop for button to be able to take in icon color if we are going to do it this way.

jaortiz: Update to new button with the 'color: inherit' so this isnt an issue anymore

PR #79, opened 2019-07-22
List

jaortiz:


brands/BOM/tokens/index.js
```diff
@@ -545,4 +545,26 @@ export default {
...
+		bullet: {
+			primary: {
+				color: COLORS.primary,
+			},
+			hero: {
+				color: COLORS.hero,
+			},
+			neutral: {
+				color: COLORS.neutral,
+			},
```
jonnystening: consider wrapping in an `appearance:` key as with other tokens?

components/list/README.md
```diff
@@ -0,0 +1,3 @@
+# @westpac/icon
+
+The icon component.
```
jonnystening: Just change to "list" (not "icon")

components/list/examples/10-appearance.js
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import { List, ListItem } from '../src';
+import { listGenerator } from './_utils';
+
+export default () => (
+	<>
+		<h3>Link List</h3>
+		<List appearance="link">
```
jonnystening: `appearance` prop should be used consistently across our components for the primary/hero/neutral/faint values. 

Perhaps we use `type` as the prop? Is type valid... since it's an html attribute?

components/list/package.json
```diff
@@ -0,0 +1,46 @@
...
+		{
+			"name": "Dominik Wilkowski",
+			"email": "Hi@Dominik-Wilkowski.com",
+			"url": "https://dominik-wilkowski.com"
+		},
+		{
+			"name": "Joss Mackison",
+			"email": "joss@thinkmill.com.au",
+			"url": "http://jossmackison.me"
+		},
```
jonnystening: I've been removing Joss from the components we've built... but left Dom in there

components/list/package.json
```diff
@@ -0,0 +1,46 @@
...
+		{
+			"name": "Jeremy Ortiz",
+			"email": "jeremy.ortiz@westpac.com.au",
+			"url": "https://jeremyortiz.io"
+		}
```
jonnystening: 👍

helpers/example/index.js
```diff
@@ -80,6 +80,11 @@ const App = ({ components, packageName }) => {
 								display: inline-block;
 								padding: 1px 3px;
 							}
+							*,
+							*:after,
+							*:before {
+								box-sizing: border-box;
+							}
```
jonnystening: Should already be in Core

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+	/** The appearance of the list */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
+	/** The color of the bullet */
+	color: PropTypes.oneOf(['primary', 'hero', 'neutral']),
+	/** The size of space between list elements */
+	size: PropTypes.oneOf(['regular', 'large']),
+	/** The icon for list */
+	icon: PropTypes.func,
+	/**  Children of List should be ListItem*/
+	children: PropTypes.oneOfType([PropTypes.arrayOf(ListItem), ListItem]),
```
jonnystening: This doesn't seem to work

Checking a component’s type doesn't seem to be recommended _ref_ appearance prop should be used consistently across our components for the primary/hero/neutral/faint values.

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+List.defaultProps = {
+	appearance: 'bullet',
+	color: 'primary',
+	size: 'regular',
```
jonnystening: use 'medium' for the default size

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+// ==============================
+// Types
+// ==============================
+List.propTypes = {
+	/** The appearance of the list */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
+	/** The color of the bullet */
+	color: PropTypes.oneOf(['primary', 'hero', 'neutral']),
+	/** The size of space between list elements */
+	size: PropTypes.oneOf(['regular', 'large']),
```
jonnystening: spacing: PropTypes.oneOf(['medium', 'large']),

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+// ==============================
+// Types
+// ==============================
+List.propTypes = {
+	/** The appearance of the list */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
+	/** The color of the bullet */
+	color: PropTypes.oneOf(['primary', 'hero', 'neutral']),
```
jonnystening: appearance: ...

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+		</ListType>
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+List.propTypes = {
+	/** The appearance of the list */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
```
jonnystening: ```type: ...```

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+		position: 'absolute',
+		left: 0,
+		top: 2,
+	};
+
+	switch (appearance) {
+		case 'bullet':
+			return (
+				<svg
+					aria-labelledby="title-bullet-list"
```
jonnystening: remove the `aria-labelledby` attribute since we aren't providing a `<title>` element

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+						strokeWidth="1"
+						fill={`${list[appearance][color].color}`}
+					/>
+				</svg>
+			);
+
+		case 'link':
+			return (
+				<svg
+					aria-labelledby="title-link"
```
jonnystening: not required

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+						fillRule="evenodd"
+						points="14.588 12 8 18.588 9.412 20 17.412 12 9.412 4 8 5.412"
+					/>
+				</svg>
+			);
+
+		case 'tick':
+			return (
+				<svg
+					aria-labelledby="title-tick-list"
```
jonnystening: not required

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+// ==============================
+// Types
+// ==============================
+ListItem.propTypes = {
+	/** The appearance of the list item */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
+	/** The color of the bullet */
+	color: PropTypes.oneOf(['primary', 'hero', 'neutral']),
+	/** The size of space between list elements */
+	size: PropTypes.oneOf(['regular', 'large']),
```
jonnystening: ```spacing: ...```

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+		</li>
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+ListItem.propTypes = {
+	/** The appearance of the list item */
+	appearance: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
```
jonnystening: ```type: ...```

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+	/** The color of the bullet */
+	color: PropTypes.oneOf(['primary', 'hero', 'neutral']),
```
jonnystening: ```appearance: ...```

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+				>
+					<circle
+						r="4"
+						cx="8"
+						cy="8"
+						stroke={`${list[appearance][color].color}`}
+						strokeWidth="1"
+						fill={`${list[appearance][color].color}`}
+					/>
+				</svg>
```
jonnystening: Do we want to consider the background image data-uri method for these? See the select element caret icons in `FormInput` component.

Keen to understand the pros/cons of each approach. There may be legacy support rationale to using background image rather than embedded svg.

jonnystening: Also.... I see you've implemented via SVG rather than css shapes. What are the pros/cons?

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+					size,
+					props,
+			  })
+			: child
+	);
+
+	const common = {
+		margin: size === 'large' ? '12px 0' : '6px 0',
+		display: 'block',
+		position: 'relative',
```
jonnystening: This better than `listStyle: 'none'` and `paddingLeft: 0` to reset?

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+
+			li: {
+				paddingLeft: 19,
+			},
+		},
+		icon: {
+			paddingLeft: 23,
+		},
+		ordered: {
+			display: 'list-item',
```
jonnystening: ref above 

"This better than `listStyle: 'none'` and `paddingLeft: 0` to reset?"

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+					css={style}
+				>
+					<circle
+						r="4"
+						cx="8"
+						cy="8"
+						stroke={`${list[appearance][color].color}`}
+						strokeWidth="1"
+						fill={`${list[appearance][color].color}`}
+					/>
```
jonnystening: With `cx="8"` and `cy="8"` and a `strokeWidth="1"` it renders a 9px diameter circle... we need 8px

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+		child && child.type && child.type === List
+			? cloneElement(child, {
+					appearance,
+					color,
+					icon,
+					size,
+					props,
+			  })
+			: child
+	);
```
jonnystening: Nested <li>'s are getting a `props="[object Object]"` attribute

components/list/src/ListItem.js
jonnystening: Just looked into this... seems the stroke is placed inside or outside depending on the browser 😬

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+
+	switch (appearance) {
+		case 'bullet':
+			return (
+				<svg
+					aria-labelledby="title-bullet-list"
+					version="1.1"
+					xmlns="http://www.w3.org/2000/svg"
+					width="16px"
+					height="16px"
```
jonnystening: SVG height and width attribute values don't require `px` units

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+					viewBox="0 0 16 16"
+					css={style}
+				>
+					<circle
+						r="4"
+						cx="8"
+						cy="8"
```
jonnystening: > "The cx and cy attributes define the x and y coordinates of the center of the circle. If cx and cy are omitted, the circle's center is set to (0,0)"

We should be able to omit cx and cy

components/list/src/ListItem.js
```diff
@@ -0,0 +1,157 @@
...
+		left: 0,
+		top: 2,
+	};
+
+	switch (appearance) {
+		case 'bullet':
+			return (
+				<svg
+					aria-labelledby="title-bullet-list"
+					version="1.1"
```
jonnystening: Remove

components/list/src/List.js
jaortiz: Will use PropTypes.node for now until I can figure out why its not working using above method or 
```
arrayOf(
		PropTypes.shape({
			type: PropTypes.oneOf([ListItem]),
		})
	),
```

brands/BOM/tokens/index.js
```diff
@@ -440,6 +440,31 @@ export default {
...
+		},
+		link: {
+			color: COLORS.primary,
+		},
+		tick: {
+			color: COLORS.primary,
+		},
+		icon: {
+			color: COLORS.primary,
```
jonnystening: should be `color: COLORS.muted`

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+	/** The size of space between list elements */
+	spacing: PropTypes.oneOf(['medium', 'large']),
+	/** The icon for list */
+	icon: PropTypes.func,
+	/**  Any renderable child */
+	children: PropTypes.node,
+};
+
+List.defaultProps = {
+	appearance: 'primary',
```
jonnystening: should be `appearance: 'hero'` according to Justin

components/list/src/ListItem.js
```diff
@@ -0,0 +1,137 @@
...
+		</li>
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+ListItem.propTypes = {
+	/** The appearance of the bullet list */
+	appearance: PropTypes.oneOf(['primary', 'hero', 'neutral']),
```
jonnystening: Do these need to be checked since the parent List already checks the props?

components/list/src/ListItem.js
```diff
@@ -0,0 +1,137 @@
...
+import { jsx, css, useTheme } from '@westpac/core';
+import svgToTinyDataURI from 'mini-svg-data-uri';
+import { List } from './List';
+
+// ==============================
+// Component
+// ==============================
+export const ListItem = ({ appearance, type, icon, spacing, children, ...props }) => {
+	const childrenWithProps = Children.map(children, child =>
+		child && child.type && child.type === List
```
jonnystening: I've also been checking with `isValidElement(child)`. Thoughts?

components/list/src/List.js
```diff
@@ -0,0 +1,47 @@
...
+	appearance: PropTypes.oneOf(['primary', 'hero', 'neutral']),
+	/** The type of the bullet */
+	type: PropTypes.oneOf(['bullet', 'link', 'tick', 'unstyled', 'icon', 'ordered']),
+	/** The size of space between list elements */
+	spacing: PropTypes.oneOf(['medium', 'large']),
+	/** The icon for list */
+	icon: PropTypes.func,
+	/**  Any renderable child */
+	children: PropTypes.node,
+};
```
jonnystening: I've been adding a note to say the prop is passed to children, so it's communicated in docz. Thoughts?

brands/BSA/tokens/index.js
```diff
@@ -440,6 +440,31 @@ export default {
...
+			color: COLORS.primary,
+		},
+		tick: {
+			color: COLORS.primary,
+		},
+		icon: {
+			color: COLORS.primary,
+		},
+	},
+
```
jonnystening: we could wrap these in `type:` to be consistent; the prop we're keying against is the type prop. Thought?

```
list: {
	type: {
		bullet: {
			appearance: {
				primary: {
					color: COLORS.primary,
				},
				hero: {
					color: COLORS.hero,
				},
				neutral: {
					color: COLORS.neutral,
				},
			},
		},
		link: {
			color: COLORS.primary,
		},
		tick: {
			color: COLORS.primary,
		},
		icon: {
			color: COLORS.primary,
		},			
	}
},
```

jaortiz: I'll change it to this for consistency but I  do think we need to discuss what our standard structure for tokens are and document it. 

i.e. whether we should be doing it
```
list[type]
```
vs
```
list.type[type]
```

I personally think that `list[type]` is nicer since we aren't repeating ourselves and reduces nesting but happy to discuss.

components/list/src/ListItem.js
jaortiz: Its still a component we are exporting so its good to have it anyway. This component can still technically work outside of our List component so good to make sure we check manual props that are passed are correct. Also doesn't really have any downsides/overhead of including it.

components/list/src/List.js
jaortiz: It looked ugly repeating for each prop so I put a message above the props table in docs saying that the styling props will be passed to children

components/list/src/ListItem.js
jaortiz: Nice find

brands/BSA/tokens/index.js
jonnystening: Other examples such as `button[appearance]` vs `button.appearance[appearance]` to be considered too

There is an added benefit to using the more verbose approach which is it often helps with spreading the CSS properties (our kind of CSS setting shorthand). But yeah definitely need to discuss decide on a token structure.

The appearance key nested inside the type key definitely feels a bit clunky

components/list/src/List.js
jonnystening: I have a pattern in forms where only a subset of the props are passed down so made sense for that. Managing this message in the component vs in docs is I guess is the thing to consider.

Let's leave what you've got and add it to discuss

components/list/src/ListItem.js
```diff
@@ -0,0 +1,138 @@
...
+				display: 'block',
+				width: 16,
+				height: 16,
+				backgroundImage: `url("${svgToTinyDataURI(linkSVG)}")`,
+				backgroundRepeat: 'no-repeat',
+				backgroundSize: 'contain',
+			},
+		},
+		tick: {
+			'&::before': {
```
jonnystening: does emotion need the `&` in the pseudo-element selector? I'm assuming it does 🤷‍♂️

components/list/src/ListItem.js
jonnystening: OK, I'll update my form component(s) so it's consistent

components/list/src/ListItem.js
jonnystening: may as well remove the `&`, emotion doesn't need them

Issue #51, opened 2019-07-10
Component: Core


jonnystening:
## Features
- [x] Global resets
  - [x] Box sizing reset
  - [x] Focus reset
  - [x] Additional resets: refer to Normalizer and Bootstrap reboot etc
- [x] Text styling
  - [x] Font styling: fontSize, fontFamily, lineHeight, color etc.
  - [x] Headings
      - [x] Focus
  - [x] Paragraphs
  - [x] Definition lists
  - [x] Abbr
  - [x] Address
  - [x] Blockquote
  - [x] Mark
  - [x] Selection styling
  - [x] Links
    - [x] Focus
- [x] Text extensions
- [x] Fonts
  - [x] Font brand & body fontFamily settings
  - [x] `@font-face` (webfont) declarations
 - [x] Focus outline styling: `isKeyboardUser` mode using react state management, provided via Context by Form component
- [x] Examples
- [ ] Jest tests
- [ ] Cypress tests
- [x] Docs

## To discuss
- [x] Breakpoints: Do we really need to be specifying a `xs: 576` breakpoint? What is that even? Isn't XS anything below the `(min-width: 768px)` breakpoint? _Note: This has been removed_

- [ ] How to include our webfonts files in our development environment?

- [x] Best way to provide 'screen reader only' functionality? Does this utility functionality just become a component in its own right? 🤔_Note: This has been moved to the `accessibility-helpers` package as the `SrOnly` component_

- [ ] Do we still need to differentiate between keyboard user navigation and mouse navigation. We are customising the focus styling, so will see keyboard outline styling appear on all UI controls (inputs, buttons, links etc). GUI2.0 uses `.is-keyboarduser` on the body to handle this. _Note: `isKeyboardUser` mode has been implemented via Context provided by Form component_

 - [ ] isKeyboardUser mode has been implemented using React state management. Should isKeyboardUser mode be managed in code via state or via body class (as per GUI2.0)? Both? How will this map across into the legacy version?

PR #87, opened 2019-07-25
Core

jonnystening:
Implementing global css resets and typography styling in the core component

Styling has been separated into resets, type, fonts and extensions for clarity; deep merged via lodash.merge

Basic text styling on the example App (`helpers/example/index.js`) has been removed

Note: This is a WIP. Refer to the component's ticket for more... _see_ https://github.com/WestpacGEL/GEL/issues/51


brands/BOM/tokens/index.js
```diff
@@ -63,7 +98,71 @@ export default {
 			foreground: 'white',
 		},
 	},
-	type: {},
+	font: FONT(),
+	typography: {
+		fontSize: 10,
```
florettewest: The base font size should be set in scalable unit (em or rem)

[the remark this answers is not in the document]
jonnystening: There's a bit of debate about this. Latest method seems to be to use fixed px for base reset and relative rems in the body. 

I'll see if I can find the articles I was referring to.

brands/BOM/tokens/index.js
```diff
@@ -273,6 +372,70 @@ export default {
...
+			},
+			invalid: {
+				borderColor: COLORS.danger,
+			},
+		},
+		size: {
+			small: {
+				padding: ['3px', '9px', '4px'],
+				fontSize: '14px',
+				height: '30px',
```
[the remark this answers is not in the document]
jonnystening: Need to discuss. Will update this on the form-input branch.

components/core/examples/10-fonts.js
```diff
@@ -0,0 +1,76 @@
...
+		<p className="body-font" style={{ fontWeight: 600 }}>
+			SemiBold: 600
+		</p>
+		<p className="body-font" style={{ fontWeight: 700 }}>
+			Bold: 700
+		</p>
+		<p className="body-font" style={{ fontWeight: 800 }}>
+			Black: 800
+		</p>
+		<p className="body-font" style={{ fontWeight: 900 }}>
```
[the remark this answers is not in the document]
jonnystening: It depends on what font the user's machine decides to use. These are just here to ensure we're checking all on all devices.

brands/BOM/tokens/index.js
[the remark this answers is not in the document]
florettewest: If you use px as a base reset, you are blocking the customisable font size in the browser setting: no good for accessibility!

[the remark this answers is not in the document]
jonnystening: Yeah I'm across this too. Not sure how what I was reading was supposed to work, it certainly doesn't resize text with a user font size adjustment 🤔. Reckon they were just ignoring these users. Not good.

css-tricks.com sets font-size: 20px for the document. Keen to know Chris' reasoning on this..

So we need to decide if we really need to have our 10px reset (using relative units). It makes rem calculations easier, but may have some drawbacks.

html { font-size: 62.5%; } /* 10px */
body { font-size: 1.4rem; } /* 14px */

or

html { font-size: 100%; } /* or not set at all, prob better */
body { font-size: 0.875rem } /* 14px */

The latter obviously means we will have a harder time working with rems, but is probably more correct for users.

brands/BOM/tokens/index.js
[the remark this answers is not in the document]
jonnystening: I've removed the height token, we can just calc this like how Bootstrap does it, and moved padding and fontSize to rems

components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,200 @@
+/** @jsx jsx */
```
jaortiz: Please be careful not to bring in other components that are not relevant to this pr.

components/core/src/Core.js
```diff
@@ -0,0 +1,296 @@
...
+			...typography.link.default,
+
+			':hover': {
+				...typography.link.hover,
+			},
+		},
+	};
+
+	// Font @fontface styling
+	const styleFont = font;
```
jaortiz: unnecessary, just use the font variable as is

components/core/src/Core.js
```diff
@@ -0,0 +1,296 @@
...
+import { paint } from './utils';
+
+import merge from 'lodash.merge';
+
+export const Core = ({ children }) => {
+	const { colors, breakpoints, font, typography } = useTheme();
+	const isKeyboardUser = useIsKeyboardUser();
+	const mq = paint(breakpoints);
+
+	console.log(`core: ${isKeyboardUser}`);
```
jaortiz: remove console.logs

components/core/src/Theme.js
```diff
@@ -19,21 +22,47 @@ export const useTheme = () => {
 	return themeObject;
 };
 
-export const GEL = ({ brand, ...props }) => {
+export const useIsKeyboardUser = () => {
```
jaortiz: update keyboardUser stuff as discussed

components/form-input/src/FormInput.js
jonnystening: will remove them 👍

jonnystening: I'll remove it for now so we can merge `core-styling` into `gel-develop`. But will have to hook this back up once we merge `form` into `gel-develop`.

components/core/src/Core.js
jonnystening: ok, will remove 'style' prefix from all emotion passed css object going forward

jaortiz: I'm fine with the rename in the destructure

PR #89, opened 2019-07-26
FormInput

jonnystening:
The component that replaces GUI2.0’s `input-fields`


components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,202 @@
...
+
+		// Disable number input spinners/steppers
+		'&::-webkit-outer-spin-button, &::-webkit-inner-spin-button': {
+			margin: 0,
+			appearance: 'none',
+		},
+	};
+
+	// Input size styling
+	const styleSize = {
```
jaortiz: Move into common, dont need a separate object

components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,202 @@
...
+		return width
+			? {
+					maxWidth: `calc(${extras}px + ${(width * factor).toFixed(1)}ex)`,
+			  }
+			: null;
+	};
+
+	// Select element styling
+	const styleSelect = () => {
+		const caretWidth = 14;
```
jaortiz: should tokenize this for consistency

components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,202 @@
...
+		const caretSVG = `<svg xmlns="http://www.w3.org/2000/svg" width="${caretWidth}" height="8" viewBox="0 0 14 8"><path fill="${
+			colors.muted
+		}" fill-rule="evenodd" d="M0 0l7 8 7-8z"/></svg>`;
+		const caretGap = formInput.select.caretGap;
+		const paddingRight = parseInt(formInput.size[size].padding[1], 10) + caretWidth + caretGap;
```
jaortiz: do the calculation in the object i.e.
`paddingRight: parseInt(formInput.size[size].padding[1], 10) + caretWidth + caretGap`

components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,202 @@
...
+			: null;
+	};
+
+	// Select element styling
+	const styleSelect = () => {
+		const caretWidth = 14;
+		const caretSVG = `<svg xmlns="http://www.w3.org/2000/svg" width="${caretWidth}" height="8" viewBox="0 0 14 8"><path fill="${
+			colors.muted
+		}" fill-rule="evenodd" d="M0 0l7 8 7-8z"/></svg>`;
+		const caretGap = formInput.select.caretGap;
```
jaortiz: unnecessary variable just use in the calculation. To help with the long variable names destructure again. i.e.
```
const { colors, formInput, typography } = useTheme();
const { select, textarea } = formInput;
```

components/form-input/src/FormInput.js
```diff
@@ -0,0 +1,202 @@
...
+	};
+
+	return (
+		<Tag
+			css={{
+				...styleCommon,
+				...styleSize,
+				...styleFixedWidth(),
+				...styleSelect(),
+				...styleTextarea,
```
jaortiz: for consistency and neatness we should be doing the select and textarea like this:
```
const styleAppearance = {
   select: {
      <styles go here>
   },
   textarea: {
      <styles go here>
   }
}
```
and then we can do `...styleAppearance[Tag]`

components/form-input/src/FormInput.js
jonnystening: thought about that, but the width value is directly related to the SVG width, which is right there...

components/form-input/src/FormInput.js
jonnystening: Sure was just keeping it consistent with some other components that separate styling relating to size, appearance etc props . But yeah not necessary

components/form-input/src/FormInput.js
jonnystening: Have had to use a css calc() for this; moved the token to rems. Removes the parseInt() which feels a bit gross anyway

Issue #80, opened 2019-07-23
Bolt new with multi-word component name


jonnystening:
When `bolt new`ing a multi-word component name the task needs to remove the dash and pascal case each of the words when it replaces the template’s `_COMPONENT_NAME_` placeholder and sets a file name.

e.g. `button-group` ==> `ButtonGroup` etc

jaortiz:
PR raised

jonnystening:
Still having issues...

Just did a `bolt new form-pod` and got this in the README.md...

```# @westpac/formPod```

Should me...

```# @westpac/form-pod```

Also in docs.mdx and package.json

`_COMPONENT_KEY_` needs to be replaced with the original CLI passed name (ie. in kebab case)

will raise a PR

jonnystening:
PR raised https://github.com/WestpacGEL/GEL/pull/92

PR #92, opened 2019-08-01
fix(cli): ensure we have correct values to replace .component-template with

jonnystening:
_ref_ https://github.com/WestpacGEL/GEL/issues/80#issuecomment-517086548

### DIFF
--- a/helpers/cli.js
+++ b/helpers/cli.js
@@ -105,15 +105,12 @@ const main = async () => {
 
 	const folderName = packageName.toLowerCase();
 
-	const compomentTitle = packageName
+	const componentTitle = packageName
 		.split('-')
 		.map(part => part[0].toUpperCase() + part.slice(1))
 		.join(' ');
 
-	const componentName = packageName
-		.split('-')
-		.map(part => part[0].toUpperCase() + part.slice(1))
-		.join('');
+	const componentName = componentTitle.replace(' ', '');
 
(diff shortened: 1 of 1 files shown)


helpers/cli.js
```diff
@@ -90,35 +90,40 @@ const main = async () => {
...
 
 	do {
 		description = await qDescription().catch(e => console.log(chalk.red(e)));
 	} while (!description);
 
-	const folderName = name.toLowerCase();
+	const folderName = packageName.toLowerCase();
 
-	name = name
+	const compomentTitle = packageName
```
jaortiz: You misspelt component :joy:

helpers/cli.js
```diff
@@ -90,35 +90,40 @@ const main = async () => {
...
+		.split('-')
+		.map(part => part[0].toUpperCase() + part.slice(1))
+		.join(' ');
+
+	const componentName = packageName
```
jaortiz: we can just do 
```
const componentName = componentTitle;
```
one less calculation

jonnystening: componentName has no spaces ("FormPod"), componentTitle has spaces ("Form Pod")

jonnystening: how about

```const componentName = componentTitle.replace(' ', '');```

helpers/cli.js
jonnystening: 😳

Issue #69, opened 2019-07-11
Component: Switch


jonnystening:
## Features
- [x] Name prop
- [x] Value prop
- [x] toggleText prop
- [x] Size prop: Small, Medium, Large, Xlarge
- [x] Block prop
- [x] Flipped prop
- [x] SrOnlyText prop
- [x] Checked prop
- [x] Disabled prop
- [x] onChange prop
- [x] Focus: `isKeyboardUser` mode via `UserModeContext` from Core
- [x] Accessibility consideration: `aria-hidden` toggle text
- [x] Examples
- [ ] Jest tests
- [ ] Cypress tests
- [x] Docs

## To discuss
- [ ] State management: Component implemented using standard CSS to style 'checked' state. Emotion is not using the checked prop to render styling. This CSS approach was taken to assist with future legacy support features.

- [ ] Auto-generate form element IDs?

- [ ] Should disabled state use a `not-allowed` cursor (as per GUI2.0)? What's best practise here?

## Feedback from Kate's review
- **Switch** - STG - Should be text colour label on Hero background, not white
_JPS: implemented in https://github.com/WestpacGEL/GEL/pull/206_

## Jonny review
- Needs focus styling
_JPS: Fixed in https://github.com/WestpacGEL/GEL/pull/206_
- Needs explicit association (for/id value) – this came back from Intopia last time

PR #104, opened 2019-08-08
Switch

jonnystening:
_Ref https://github.com/WestpacGEL/GEL/issues/69_


components/form-switch/src/FormSwitch.js
```diff
@@ -0,0 +1,165 @@
...
+
+	/**
+	 * The form switch input element’s id.
+	 *
+	 * This prop is required.
+	 */
+	id: PropTypes.string.isRequired,
+
+	/**
+	 * The form switch on/off text.
```
florettewest: Could be more precise on how to change the text?

components/form-switch/examples/00-default.js
```diff
@@ -0,0 +1,48 @@
...
+		</FormSwitch>
+		<FormSwitch name="ex3" id="ex3" toggleText={['Yes', 'No']} srOnlyText>
+			Screen reader only text
+		</FormSwitch>
+
+		<hr />
+
+		<h3>Switches with text label</h3>
+		<FormSwitch name="ex4" id="ex4">
+			Enable notifications
```
florettewest: For all "Enable notifications" label, the switch text could be "Yes/No"

components/form-switch/examples/10-sizes.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+
+import { FormSwitch } from '../src';
+
+export default () => (
+	<>
+		<h3>Small</h3>
+		<FormSwitch name="ex1" id="ex1" size="small">
+			Enable notifications
```
florettewest: For all "Enable notifications" label, the switch text could be "Yes/No"

components/form-switch/examples/20-options.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+import { FormSwitch } from '../src';
+
+export default () => (
+	<>
+		<h3>Flipped switches</h3>
+		<FormSwitch name="ex1" id="ex1" flip>
+			Enable notifications
```
florettewest: For all "Enable notifications" label, the switch text could be "Yes/No"

components/form-switch/examples/30-responsive.js
```diff
@@ -0,0 +1,28 @@
...
+
+import { FormSwitch } from '../src';
+
+export default () => (
+	<>
+		<h3>This breakpoint and wider</h3>
+		<FormSwitch name="ex1" id="ex1" size={['medium', 'xlarge']} block>
+			Extra large for MD breakpoint and wider
+		</FormSwitch>
+		<FormSwitch name="ex2" id="ex2" size={['small', 'small', 'xlarge']} block>
```
florettewest: Should we say "null" for consistency or leave it as "small" to show it works both ways?
size={['small', null, 'xlarge']}

components/form-switch/src/FormSwitch.js
```diff
@@ -0,0 +1,165 @@
...
+					position: 'absolute',
+					zIndex: -1,
+					opacity: 0,
+				}}
+				name={name}
+				id={id}
+				defaultChecked={checked}
+				disabled={disabled}
+			/>
+			{children && (
```
florettewest: Do we need to test if there is children if you have it set as "required" in the propTypes? 
children: PropTypes.string.isRequired,

components/form-switch/docs.mdx
```diff
@@ -0,0 +1,18 @@
...
+
+import { Playground, PropsTable } from 'docz';
+import { Component } from './src';
+
+# FormSwitch
+
+Write your docs here
+
+<Playground>
+	<FormSwitch />
```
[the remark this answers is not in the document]
jonnystening: I haven't done docs for any of my components yet. Not sure if we should now, or after our thing with TM. Prob should now I guess

components/form-switch/examples/00-default.js
jonnystening: You're thinking about automating it? Enable... gets a yes/now. Turn... gets a on/off?

I think its best we don't lock any of this down. Let the dev implement what is required by the project on their end. We don't know what they're going to need and don't want to get in their way.

jonnystening: Ooooooh are you just saying for the example? Yeah that's a good idea :)

components/form-switch/src/FormSwitch.js
jonnystening: Good pick up! 

I'm not sure. It can't hurt to have it checked twice I guess. Can we rely solely on prop types set to isRequired? That's a TM question

jonnystening: Remember propTypes only warn, they don't stop things from breaking.

components/form-switch/examples/00-default.js
florettewest: Yes just for example... and it kind of makes sense if you answer the question: "Do you want to enable the notifications?" Yes or No... On or Off seems weird in that context...

components/switch/src/styled.js
```diff
@@ -0,0 +1,147 @@
...
+export const SwitchText = ({ size, block, flip, srOnlyText, ...props }) => {
+	const { breakpoints, switch: formSwitch } = useTheme();
+	const mq = paint(breakpoints);
+
+	const styleText = {
+		flex: block ? 1 : null,
+		display: 'flex',
+		alignItems: 'center',
+		minHeight: asArray(size).map(s => s && formSwitch.size[s].height),
+		...(gap => (flip ? { paddingLeft: gap } : { paddingRight: gap }))(formSwitch.text.gap),
```
jaortiz: dont see the value in having this as an IIFE. Should just be :
```
...(flip ? { paddingLeft: formSwitch.text.gap } : { paddingRight: formSwitch.text.gap }),
```

components/switch/src/styled.js
```diff
@@ -0,0 +1,147 @@
...
+			height: sizeArr.map(s => s && formSwitch.size[s].height),
+			fontSize: sizeArr.map(s => s && formSwitch.size[s].fontSize),
+		};
+	};
+
+	// On/off toggle switch styling
+	const styleToggle = {
+		borderRadius: styleResponsive().borderRadius,
+		width: styleResponsive().width,
+		height: styleResponsive().height,
```
jaortiz: For optimisation call the styleReponsive function once and assign to a variable. i.e. 
```
const responsive = styleResponsive();

borderRadius: responsive.borderRadius
```
(feel free to use whatever name)

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+
+export const Switch = ({
+	name,
+	id,
+	toggleText,
+	size,
+	block,
+	flip,
+	srOnlyText,
+	defaultChecked,
```
jaortiz: should be checked since defaultChecked will only set the initial value and leave it up to the html element to manage its internal state (uncontrolled component) which we dont want.

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+	disabled,
+	children,
+	...props
+}) => {
+	const { switch: formSwitch } = useTheme();
+
+	const [isChecked, setIsChecked] = useState(defaultChecked);
```
jaortiz: update to useState(checked)

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+			<input
+				type="checkbox"
+				css={{
+					position: 'absolute',
+					zIndex: -1,
+					opacity: 0,
+				}}
+				name={name}
+				id={id}
+				defaultChecked={defaultChecked}
```
jaortiz: update to `checked={isChecked}`

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+	defaultChecked,
+	disabled,
+	children,
+	...props
+}) => {
+	const { switch: formSwitch } = useTheme();
+
+	const [isChecked, setIsChecked] = useState(defaultChecked);
+	const toggle = () => setIsChecked(!isChecked);
+
```
jaortiz: Since we want this to act as a controlled component we should be mapping the current checked state to the app state i.e. should be dependent on the checked prop passed in. Need to add useEffect hook to update our state when new props are passed in
```
useEffect(
		() => {
			setIsChecked(checked);
		},
		[checked]
	);
```

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+export const Switch = ({
+	name,
+	id,
+	toggleText,
+	size,
+	block,
+	flip,
+	srOnlyText,
+	defaultChecked,
+	disabled,
```
jaortiz: we should also be taking an onToggle prop to allow for the app to control the state of this switch (controlled components paradigm)

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+		width: block ? '100%' : null,
+		flexDirection: flip ? 'row-reverse' : null,
+
+		':hover': {
+			cursor: 'pointer',
+		},
+	};
+
+	return (
+		<label htmlFor={id} css={common} onChange={toggle} {...props}>
```
jaortiz: onChange should be on the input element

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+	srOnlyText,
+	defaultChecked,
+	disabled,
+	children,
+	...props
+}) => {
+	const { switch: formSwitch } = useTheme();
+
+	const [isChecked, setIsChecked] = useState(defaultChecked);
+	const toggle = () => setIsChecked(!isChecked);
```
jaortiz: This should ideally be the onToggle function passed in but in case they dont pass on I suggest doing
```
const toggle = () => {
		if (onToggle) {
			onToggle();
		} else {
			setIsChecked(!isChecked);
		}
	};
```
Need to discuss further with thinkmill about input component control patterns

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+		<label htmlFor={id} css={common} onChange={toggle} {...props}>
+			<input
+				type="checkbox"
+				css={{
+					position: 'absolute',
+					zIndex: -1,
+					opacity: 0,
+				}}
+				name={name}
+				id={id}
```
jaortiz: We should auto-gen this id. Is there ever a case for them to pass in their own id? Would ensure we are always accessible

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+Switch.propTypes = {
+	/**
+	 * Input element name attribute.
+	 */
+	name: PropTypes.string,
+
+	/**
+	 * Input element id attribute.
+	 *
+	 * This prop is required.
```
jaortiz: Unnecessary as docz documents this for us via the props.isRequired

components/switch/src/switch.js
```diff
@@ -0,0 +1,165 @@
...
+	 * On/off text.
+	 *
+	 * This prop takes an array where the first index is the "on" text and second index is the "off" text e.g. "['Yes', 'No']"
+	 */
+	toggleText: PropTypes.arrayOf(PropTypes.string),
+
+	/**
+	 * Form switch size.
+	 *
+	 * Defaults to "medium"
```
jaortiz: Unnecessary as docz documents this via defaultProps

components/switch/src/styled.js
```diff
@@ -0,0 +1,147 @@
...
+		position: 'relative',
+		zIndex: 1,
+		border: `${formSwitch.borderWidth}px solid`,
+		borderColor: formSwitch.borderColor,
```
jaortiz: combine in one line i.e.
```
border: `${formSwitch.borderWidth}px solid ${formSwitch.borderColor}`
```

components/switch/src/switch.js
jonnystening: Whoops

components/switch/src/switch.js
jonnystening: Not sure

Is it best practise to auto-generate ID values within the component?

Can we mark this for discussion with TM?

jonnystening: Found this... 

https://codeburst.io/unique-ids-and-where-to-find-them-14020ddbc964

jaortiz: Yea we can mark as discussion. I like autogenerating as it ensures that there is always a unique id and makes one less thing for the dev to worry about.

Example in modal
https://github.com/WestpacGEL/GEL/blob/689c5a5061fc24d0d33287a76f7ab246714dcbdb/components/modal/src/Modal.js#L35

jonnystening: I've implemented auto-generate as above :)

Issue #64, opened 2019-07-11
Component: Symbol


jonnystening:
## Features
- [x] Width prop
- [x] Height prop
- [x] viewBoxWidth prop
- [x] viewBoxHeight prop
- [x] Symbols
- [x] Logos: Default and multi-brand
- [ ] Accessibility features: aria-label or sr-only? 🤔
- [x] Examples
- [ ] Jest tests
- [ ] Cypress tests
- [x] Docs

## Issues
- [ ] Not rendering props table in docs. Not sure why 🤷‍♂️

## To discuss
- [ ] How do we feel about having GUI2.0's logos and symbols merged into this one component?

- [ ] Is ‘Symbol’ the best name for this component?

- [ ] As with Icon (https://github.com/WestpacGEL/GEL/issues/48)... is `aria-label` supported well enough by screen readers? Intopia have previously told us to use ‘sr-only’ technique for alt text

- [ ] I've stored the multi-brand logo's SVG code as strings in a new SYMBOL token. To render this in the multi-brand components (LogoLarge & LogoSmall) I've used `dangerouslySetInnerHTML` on a svg group `<g>` tag, since that's the only way I could work out how to do it. Keen to learn the right way 🤔

- [ ] After some thinking (I sometimes do this when running or in the shower)... perhaps the resize feature (`width` and `height` props) is something that Symbol should _not_ be concerned with. We allow sizing of the Icon component via size prop, but that's because we mandate a set of sizes. The sizing of Symbols can be done externally I now reckon.

## Feedback from Kate's review
- **symbols/all symbols** -  _WindowsStoreSymbol_ missing
_JPS: Broken, but needs to be removed and replaced with the Microsoft store badge logo_
_JPS: WindowsStoreSymbol logo removed_

## Jonny review
- Missing multi-brand logos

PR #105, opened 2019-08-09
Symbol

jonnystening:
Ref https://github.com/WestpacGEL/GEL/issues/64


components/symbol/src/Symbol.js
```diff
@@ -0,0 +1,117 @@
...
+		verticalAlign: 'middle',
+	};
+
+	// Reponsive styling (symbol size)
+	const styleResponsive = () => {
+		const widthArr = asArray(width || viewBoxWidth);
+		const heightArr = asArray(height || viewBoxHeight);
+
+		const ratio = viewBoxWidth / viewBoxHeight;
+		const resizeByWidth = !!width;
```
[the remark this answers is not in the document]
jonnystening: bang bang... evaluates to boolean true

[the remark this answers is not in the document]
florettewest: that shot me down... 
Thanks

components/symbol/src/Symbol.js
```diff
@@ -0,0 +1,115 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useTheme, paint } from '@westpac/core';
+
+// ==============================
+// Utils
+// ==============================
+
+const asArray = val => (Array.isArray(val) ? val : [val]);
+const round = f => Math.round(f * 100) / 100; //2DP
```
jaortiz: is this necessary? what is the convention/best practise for rounding css values

components/symbol/src/Symbol.js
```diff
@@ -0,0 +1,115 @@
...
+		verticalAlign: 'middle',
+	};
+
+	// Reponsive styling (symbol size)
+	const styleResponsive = () => {
+		const widthArr = asArray(width || viewBoxWidth);
+		const heightArr = asArray(height || viewBoxHeight);
+
+		const ratio = viewBoxWidth / viewBoxHeight;
+		const resizeByWidth = !!width;
```
jaortiz: Unnecessary, just use width as is to determine truthy/falsy

components/symbol/src/Symbol.js
jonnystening: Very hard to find a definitive answer 🤔

The rounding of fractional pixel values is not consistent across browsers. Some will round down (floor), some up (ceil) and some truncate decimal places.

My thinking was we should set it as accurate as possible; let the browser decide, but 1DP is probably all that's required if we do keep decimal places.

We can probably just use Math.round() in this case to return a 0DP pixel value. This will skew resized symbols/logos ever so slightly; maybe only noticeable on retina monitors, if at all.

@florettewest have you had to deal with this previously?

florettewest: @jonnystening, I am for rounding to 1 or 2DP. Because it is more accurate, and that's the whole point of using SVGs. Browsers might align in the future?

PR #108, opened 2019-08-13
ListGroup

florettewest:


components/list-group/src/ListGroup.js
```diff
@@ -0,0 +1,48 @@
+/** @jsx jsx */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useTheme } from '@westpac/core';
+
+// ==============================
+// Utils
+// ==============================
```
jaortiz: remove comment as unused

components/list-group/src/ListGroupItem.js
```diff
@@ -0,0 +1,47 @@
+/** @jsx jsx */
+
+import React from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useTheme } from '@westpac/core';
+
+// ==============================
+// Utils
```
jaortiz: remove as unused

components/list-group/examples/_utils.js
```diff
@@ -0,0 +1,7 @@
+/** @jsx jsx */
+
```
jaortiz: just move this into the 00-default.js file since its small

PR #112, opened 2019-08-16
Convert all units to rems

jonnystening:


florettewest:
I suggest to extend the use of rems to border-radius: it will keep the look of "pills" elements.
Badge currently without rems border-radius: 
Screen Shot 2019-08-16 at 3 35 10 pm

components/list/src/ListItem.js
```diff
@@ -25,9 +25,9 @@ export const ListItem = ({ appearance, type, icon, spacing, children, ...props }
 	const { list } = useTheme();
 
 	const common = {
-		margin: spacing === 'large' ? '12px 0' : '6px 0',
+		margin: spacing === 'large' ? '1.2prem 0' : '0.6rem 0',
```
florettewest: 1.2prem

jonnystening: Good pick up @florettewest :)

PR #115, opened 2019-08-16
Icon tweak

jonnystening:


components/icon/src/Icon.js
```diff
@@ -33,7 +33,7 @@ const IconWrapper = ({ size, ...props }) => {
 
 	// Reponsive styling (icon size)
 	const styleResponsive = () => {
-		const sizeArr = asArray(size).map(s => sizeMap[s]);
+		const sizeArr = asArray(size).map(s => (s !== null ? sizeMap[s] : null));
```
jaortiz: remove `!== null` as it is unecessary, null is falsy

jonnystening: Facepaint doesn't mind receiving null or false, it will skip returning a media query for that breakpoint (as desired). 

> "Boolean, undefined, and null values are ignored." 
— https://github.com/emotion-js/facepaint

Without the null check... `size={['small', null, 'large']}` will return `['small', false, 'medium']` in the case above.

I think this check is more of a thing when we're passing non-standard facepaint array values such as for Button's responsive block (which actually takes true/false), or possibly a prop that can take a zero value (since 0 is falsy).

I've just brought the logic across for all of these responsive styling blocks for consistency, but yeah probably overkill.

jonnystening: have pushed commit to simplify this to a `s && sizeMap[s]`

Added a null example too

PR #130, opened 2019-09-10
InputGroup

florettewest:


brands/BOM/tokens/index.js
```diff
@@ -676,6 +676,39 @@ export default {
 		},
 	},
 
+	inputGroup: {
+		addon: {
+			fontWeight: '400',
```
jonnystening:
```suggestion
			fontWeight: 400,
```

brands/BOM/tokens/index.js
```diff
@@ -676,6 +676,39 @@ export default {
...
+			lineHeight: '1.5',
```
jonnystening:
```suggestion
			lineHeight: 1.5,
```

components/input-group/README.md
```diff
@@ -0,0 +1,5 @@
+# @westpac/input-group
+
+Styled input fields with attached addons. Addons can be text (\$, %, .00 etc) or form controls (buttons or select inputs).
```
jonnystening: \$ ?

Remove the slash

components/input-group/docs.mdx
```diff
@@ -0,0 +1,92 @@
...
+---
+
+import { Playground, PropsTable } from 'docz';
+import { InputGroup, InputGroupAddon } from './src';
+import { Button } from '../button/src';
+import { TextInput } from '../text-input/src';
+
+# InputGroup
+
+Styled input fields with attached addons. Addons can be text (\$, %, .00 etc) or form controls (buttons or select inputs).
```
jonnystening: Remove `\$` slash

components/input-group/examples/_utils.js
```diff
@@ -0,0 +1,13 @@
...
+
+export const Box = props => (
+	<div
+		css={{
+			alignItems: 'center',
+			display: 'flex',
+		}}
+		{...props}
+	/>
+);
```
jonnystening: Remove this file if it's not being used

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+			marginLeft: '-1px',
+		},
+
+		'select:not(:last-child)': {
+			marginRight: '-1px',
+		},
+
+		'input:not(:first-child), span:not(:first-child), button:not(:first-child), select:not(:first-child)': {
+			borderTopLeftRadius: '0',
+			borderBottomLeftRadius: '0',
```
jonnystening: Just make 0s integer values

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+			borderBottomRightRadius: '0',
+		},
+
+		'input + span': {
+			borderTopRightRadius: '3px',
+			borderBottomRightRadius: '3px',
+		},
+
+		'span:first-child, button:first-child': {
+			borderRight: 'none',
```
jonnystening: Use 0 rather than 'none'

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+		},
+
+		'span:last-child, button:last-child': {
+			borderLeft: 'none',
```
jonnystening: Use 0 rather than 'none'

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+	 * Button appearance
+	 */
+	appearance: PropTypes.oneOf(options.appearance),
+
+	/**
+	 * Button tag
+	 */
+	tag: PropTypes.oneOf(options.tag),
```
jonnystening: This isn't a prop of InputGroup's

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+};
+
+// ==============================
+// Types
+// ==============================
+
+const options = {
+	appearance: ['primary', 'hero', 'neutral', 'faint', 'link'],
+	size: ['small', 'medium', 'large', 'xlarge'],
+	tag: ['select'],
```
jonnystening: not required, see below

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+InputGroup.propTypes = {
+	/**
+	 * InputGroup size
+	 */
+	size: PropTypes.oneOf(options.size),
+
+	/**
+	 * Button appearance
+	 */
+	appearance: PropTypes.oneOf(options.appearance),
```
jonnystening: Not sure we should be setting 'appearance' of the internal buttons at a InputGroup level. Size makes sense, since it's passed to all children ensuring consistency. The appearance prop can be given to any buttons individually as required.

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+// ==============================
+// Component
+// ==============================
+
+/**
+ * Input Group: Styled input fields with attached addons. Addons can be text ($, %, .00 etc) or form controls (buttons or select inputs).
+ */
+export const InputGroup = ({ size, appearance, children, ...props }) => {
+	const childrenWithProps = Children.map(children, child =>
+		cloneElement(child, { size, appearance })
```
jonnystening: I think size would be worthwhile, but appearance should be set per button. See comment in PropTypes below.

Thing is that size is also passed to the components via form, consumed from FormContext (assuming all form elements in a form should be the same size). So it's more likely they will receive a size value this way.

```
<Form size="large">
   <InputGroup>
      <TextInput />
      <Button>Go</Button>
   </InputGroup
</Form>
```

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+	tag: PropTypes.oneOf(options.tag),
+
+	/**
+	 * InputGroup children
+	 */
+	children: PropTypes.node.isRequired,
+};
+
+InputGroup.defaultProps = {
+	size: 'medium',
```
jonnystening: not required based on comments above

components/input-group/src/InputGroupAddon.js
```diff
@@ -0,0 +1,46 @@
...
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+
+InputGroupAddon.propTypes = {
+	/**  Any renderable child */
+	children: PropTypes.node,
```
jonnystening: add size propType

components/input-group/src/InputGroupAddon.js
```diff
@@ -0,0 +1,46 @@
...
+};
+
+InputGroupAddon.defaultProps = {};
```
jonnystening: set a size defaultProps

components/input-group/src/InputGroupAddon.js
```diff
@@ -0,0 +1,46 @@
...
+		</span>
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+
+InputGroupAddon.propTypes = {
+	/**  Any renderable child */
```
jonnystening: Could probably re comment this as 'Component text'

components/input-group/src/InputGroupAddon.js
jonnystening:
```suggestion
	children: PropTypes.string.isRequired,
```

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,99 @@
...
+		</div>
+	);
+};
+
+// ==============================
+// Types
+// ==============================
+
+const options = {
+	appearance: ['primary', 'hero', 'neutral', 'faint', 'link'],
```
jonnystening: Note required based on comment above

components/input-group/README.md
florettewest: It is markdown escape...

components/input-group/src/InputGroup.js
```diff
@@ -0,0 +1,85 @@
...
+
+		'select:not(:first-child)': {
+			marginLeft: '-1px',
+		},
+
+		'select:not(:last-child)': {
+			marginRight: '-1px',
+		},
+
+		'input:not(:first-child), span:not(:first-child), button:not(:first-child), select:not(:first-child)': {
```
jonnystening: should change the `:first-child` selectors to `:first-of-type` selectors as per react server side rendering recommendation

components/input-group/docs.mdx
jonnystening: Screen Shot 2019-09-11 at 4 01 29 pm

components/input-group/docs.mdx
```diff
@@ -0,0 +1,92 @@
...
+## Button
+
+<Playground>
+	<InputGroup appearance="hero">
+		<TextInput />
+		<Button>Go</Button>
+	</InputGroup>
+</Playground>
+
+## select
```
jonnystening:
```suggestion
## Select
```

components/input-group/docs.mdx
```diff
@@ -0,0 +1,92 @@
+---
+name: InputGroup
+menu: Components
+route: /input-group
+---
+
+import { Playground, PropsTable } from 'docz';
+import { InputGroup, InputGroupAddon } from './src';
+import { Button } from '../button/src';
```
jonnystening:
```suggestion
import { Button } from '@westpac/button';
```

components/input-group/docs.mdx
```diff
@@ -0,0 +1,92 @@
...
+import { TextInput } from '../text-input/src';
```
jonnystening:
```suggestion
import { TextInput } from '@westpac/text-input';
```

components/input-group/src/InputGroup.js
florettewest: `:first-of-type` does not work in that case. I need to find another way (not css selector) to target these elements...
We do not use SSR so should we ignore? https://github.com/emotion-js/emotion/issues/1105

components/input-group/docs.mdx
florettewest: I cannot remove the / 🤔
Prettier put it back...

components/input-group/src/InputGroup.js
```diff
@@ -12,7 +13,17 @@ import { jsx } from '@westpac/core';
  * Input Group: Styled input fields with attached addons. Addons can be text ($, %, .00 etc) or form controls (buttons or select inputs).
  */
 export const InputGroup = ({ size, children, ...props }) => {
-	const childrenWithProps = Children.map(children, child => cloneElement(child, { size }));
+	const formContext = useContext(FormContext);
+
+	console.log('context', formContext);
+
+	const inputGroupSize = formContext.size || size;
```
jonnystening: I think we should do this the other way around so the prop has precedence 

`const inputGroupSize = size || formContext.size || 'medium';

You will have to remove the default prop down below too

components/input-group/src/InputGroup.js
```diff
@@ -12,7 +13,17 @@ import { jsx } from '@westpac/core';
...
+
+	console.log('inputGroupSize', inputGroupSize);
+
+	const childrenWithProps = Children.map(children, child =>
+		cloneElement(child, { size: inputGroupSize })
+	);
```
jonnystening: This is replaced with the context approach, so you don't need to cloneElement() any more

jonnystening:
I've had a go at this component, taking learnings from some others we've been refactoring.

Learnings...
 - Use context to pass props to children (don't prop-drill)
 - But it's okay to use prop-drilling/cloneElement() pattern to pass simple first/last props, used for styling
 - You can create special use-case versions of existing components to suit a particular component's needs by wrapping (kind of like extending)

Based on these considerations, recent commits to this branch include...

 - Merge updated branches required to render an InputGroup (`gel-develop`, `button-group/button/icon`, `form/text-input`)
 - InputGroup creates its own context (InputGroupContext) to pass `size` to children
 - InputGroup also consumes FormContext (which may be available via a wrapping Form component setting a default size for the form) — as per other form components which also consume FormContext
 - InputGroup sub-components created to wrap existing components (Button and TextInput) and have CSS overrides to style internal borders etc. These CSS overrides were on the parent (InputGroup), now moved to each individual sub-component.
 - InputGroup iterates its children, passing `first` and `last` props via cloneElement(), depending on child index. The children component's styling is provided depending on these props. 
- We export nice/short component names to improve DX

Think that's about it 😀

florettewest:
This is not relevant anymore... I will open a PR on the updated InputGroup component.

Issue #41, opened 2019-07-10
Component: ButtonGroup


jonnystening:
## Features
- [x] Appearance prop
- [x] Size prop
- [x] Block prop
- [x] Name prop
- [x] Icons
- [ ] Button group text (similar to switch label) – update ListGroup example once implemented
- [x] Examples
- [ ] Jest tests
- [ ] Cypress tests
- [x] Docs

## Sub-components
- [x] ButtonGroupButton
  - [x] Value prop
  - [x] iconAfter prop
  - [x] iconBefore prop
  - [x] SrOnlyText prop
  - [x] Checked prop
  - [x] onChange prop

## To discuss
- [ ] Currently setting button's borderRadius and border styling with CSS. Consider doing this programmatically, iterating children and applying borders as needed? Similar to InputGroup question.

- [ ] We should probably limit the button appearance options. Do we even allow other appearances or is hero appearance it? 

- [ ] Had a thought. This could really easily be merged into the FormCheck component; provided via prop type="button" (adding to radio and checkbox types). Thoughts? @simonswiss @dominikwilkowski 

## Feedback from Kate's review
- **Button group** -  See Buttons #40 

## Jonny review
- Missing hidden radio inputs

PR #134, opened 2019-09-24
Example of context API instead of React.cloneElement

simonswiss:
Why: using React.cloneElement breaks out of the composition model of React, and can lead to some errors if a developer ads/remove a wrapping `<div>` for example.

This PR is an example of how this could be replaced with the Context API in the example of the `ButtonGroup` component.


components/button-group/src/ButtonGroupButton.js
```diff
@@ -23,6 +22,8 @@ export const ButtonGroupButton = ({
 	children,
 	...props
 }) => {
+	const { appearance, size, name } = useButtonGroupContext();
```
mitchellhamilton: Shouldn't name always be set on a per-button basis?

Wouldn't you also want to be able customize appearance on a per button basis? E.g. One button is primary, one neutral, etc.

Size probably shouldn't be customizable on a per button basis though because that would look pretty strange.

simonswiss: These are good questions worth a discussion with design/UX team.

For now I haven't changed the current behaviour, where these are set on the `ButtonGroup` level.

mitchellhamilton: This change would mean that name and appearance couldn't be set per button which previously wasn't true though.

simonswiss: Update: I had a phone conversation with @jonnystening - and the intent here is to actually _prevent_ users from changing the appearance / name / size from within the children `ButtonGroupButton`.

jonnystening: Yeah re above... Our ButtonGroup is really just a radio toggle set. We wouldn't have a requirement for a ButtonGroup comprising of different button appearances, as with size. The button children are always the same appearance and size. 

Name prop too... as a radio they need to always have the same name for the html radios beneath to function.

jonnystening: Note: our ButtonGroup is different to AtlasKit's. Theirs is a button wrapper that just sets a gap between inline buttons. Ours is really just a radio set (one toggle-able at at time), but with a button appearance.

Which takes me to something I discussed with @simonswiss yesterday... We have a component called FormCheck (name came from what Bootstrap did a while back). 

See GUI2.0's (Form-check)[https://westpaccxteam.github.io/GUI-form-checks/tests/WBC/]. I built this for GEL3, it's in the gel-develop branch.

A FormCheck is a set of radios or checkboxes. It takes a type prop (type="radio" or "checkbox"). This component renders custom toggle indicators using CSS shapes (which have been though a11y testing etc). 

As mentioned on the (ButtonGroup ticket)[https://github.com/WestpacGEL/GEL/issues/41]. It could be possible to merge the functionality of ButtonGroup into FormCheck since under the hood `<ButtonGroup />` is really the same as a `<FormCheck type="radio" />`. A new `<FormCheck type="button" />` prop value could provide this. 

I would just need to learn how to dynamically import the @westpac/button package when rendering that button mode, and would also mean the FormCheck would need to accept an appearance prop (but only used for rendering button mode).

🤷‍♂️

PR #135, opened 2019-09-25
Remove document 10px fontSize reset

jonnystening:
Reverts the document back to a 16px font-size default

1rem = 16px


jonnystening:
Yeah sorry... it's branched from @simonswiss's ButtonGroup example. Wasn't mean to be. Will try again 🙄

PR #136, opened 2019-09-25
Update ButtonGroup, Button and Icon

jonnystening:
## Updates
- Fixed styling in our new more better-er token world (Button & ButtonGroup)
- Dropped the isPropname syntax (e.g. `isBlock` now `block`)
- Use Context rather than prop-drill children, as per Simon's example
- Export ButtonGroupButton as Button
- Fixes a minor example frame sidebar styling issue, better aligns all items on the left

## Still to fix...
- How to pass a checked button state (ie. set a pre-checked button to show immediately on page load)
- STG brand requires a unique colour mapping (must use `COLORS.text` on top of `COLORS.hero`). This is something we can implement when we bring back the default/foreground colour token feature.


jonnystening:
@jossmac your ‘por que no los dos’ implementation approach is 👌

PR #138, opened 2019-09-26
Modal

jaortiz:
Update for new tokens, use context now, removed modal prefix from sub components


components/modal/src/Modal.js
```diff
@@ -1,15 +1,35 @@
 /** @jsx jsx */
 
-import React, { Children, cloneElement, useState, useEffect } from 'react';
+import React, {
+	Children,
+	cloneElement,
```
jonnystening: Children and cloneElement now no longer used

components/modal/src/Modal.js
```diff
@@ -1,15 +1,35 @@
...
+// ==============================
+// Context and Consumer Hook
+// ==============================
+const ModalContext = createContext();
+
+export const useModalContext = () => {
+	const context = useContext(ModalContext);
+
+	if (!context) {
+		throw new Error('Modal inner components should be wrapped in a <Modal>.');
```
jonnystening: Should make this message consistent

```suggestion
		throw new Error('Modal sub-components should be wrapped in a <Modal>.');
```

?

components/modal/src/styled.js
```diff
@@ -36,43 +37,45 @@ export const ModalBackdrop = props => {
 };
 
 export const StyledModal = ({ size, ...props }) => {
-	const { modal, breakpoints } = useTheme();
+	const {
+		LAYOUT: { breakpoints },
+	} = useTheme();
 	const mq = paint(breakpoints);
 
 	const styles = {
```
jonnystening: Inline these

components/modal/src/styled.js
```diff
@@ -3,7 +3,7 @@
 import React from 'react';
 import { jsx, useTheme, paint } from '@westpac/core';
 
-export const ModalBackdrop = props => {
+export const Backdrop = props => {
 	const styles = {
```
jonnystening: Inline these

components/modal/src/styled.js
```diff
@@ -36,43 +37,45 @@ export const ModalBackdrop = props => {
...
 export const Title = props => {
-	const { modal } = useTheme();
+	const { COLORS } = useTheme();
 	return (
 		<span
 			css={{
-				fontSize: modal.title.fontSize,
-				fontWeight: modal.title.justifyContent,
-				color: modal.title.color,
+				fontSize: 18,
```
jonnystening: rem value

components/modal/src/styled.js
```diff
@@ -36,43 +37,45 @@ export const ModalBackdrop = props => {
...
+		backgroundColor: '#fff',
+		borderRadius: 3,
 		boxShadow: '0 5px 15px rgba(0,0,0,0.5)',
 		transition: 'all 0.3s ease',
 		width: [
 			'auto',
-			...(size === 'small' ? [modal.size.small.width] : [modal.size.medium.width]),
-			...(size === 'large' ? [modal.size.large.width] : []),
+			...(size === 'small' ? ['18.75rem'] : ['37.5rem']),
+			...(size === 'large' ? ['56.25rem'] : []),
```
jonnystening: Do you need to pass arrays?

components/modal/examples/10-size.js
```diff
@@ -25,33 +25,31 @@ export default () => {
...
-				</ModalFooter>
+					</button>
+				</Footer>
 			</Modal>
 
 			<p>
-				<Button onClick={() => updateModal('small', true)}>Open small modal</Button>
+				<button onClick={() => updateModal('small', true)}>Open small modal</button>
 			</p>
 			<Modal isOpen={state.small.open} onClose={() => updateModal('small', false)} size="small">
```
jonnystening:
```suggestion
			<Modal open={state.small.open} onClose={() => updateModal('small', false)} size="small">
```

etc...

components/modal/src/Modal.js
```diff
@@ -1,20 +1,33 @@
...
+};
 
 // ==============================
 // Component
 // ==============================
-export const Modal = ({ isOpen, onClose, size, children, ...props }) => {
+export const Modal = ({ open: isOpen, onClose, size, children, ...props }) => {
 	const [open, setOpen] = useState(isOpen);
 
 	useEffect(() => {
```
jonnystening: Hey shouldn't refer to the `open` state, rather than the `isOpen` prop?

```
useEffect(
		() => {
			setOpen(open);
		},
		[open]
	);
```

components/modal/src/ModalHeader.js
```diff
@@ -20,36 +21,26 @@ export const ModalHeader = ({ id, onClose, children, ...props }) => {
...
+				borderBottom: `1px solid ${COLORS.border}`,
+				padding: '1rem 1.5rem 0.75rem',
 			}}
 			{...props}
 		>
-			<Title id={id}>{children}</Title>
-			<Button
+			<Title id={titleId}>{children}</Title>
+			{/* waiting until button component is fixed */}
+			{/* <Button
```
jonnystening: can reimplement this now...

components/modal/src/Modal.js
jaortiz: No, since Im running the effect to update the internal state if the isOpen prop that is passed in changes. The modal should be controlled by the props being passed into it.

PR #140, opened 2019-09-26
Update Panel

jonnystening:
Updates styling to use new tokens

Uses Context, removes pop-drilling

Removes sub-component props and PropType checking

Exports sub-components without parent component prefix (e.g. PanelHeading --> Heading)


components/panel/src/PanelBody.js
```diff
@@ -2,23 +2,16 @@
...
 // Component
 // ==============================
 
-export const PanelBody = ({ appearance, ...props }) => {
-	const { breakpoints, panel } = useTheme();
+export const PanelBody = props => {
+	const {
+		LAYOUT: { breakpoints },
+	} = useTheme();
 	const mq = paint(breakpoints);
```
jossmac: Not directly relate to this PR, but it seems odd that we're passing `paint` around and creating the helper utility inside each component. Might be a good time to abstract it e.g.

```jsx
// Somewhere in Core?
export const useMediaQuery = () => {
  const { LAYOUT } = useTheme();

  return paint(LAYOUT.breakpoints);
}

// Use
export const Well = props => {
  const mq = useMediaQuery();

  return (
    <div css={mq({
      query: 'reliant',
      styles: ['can', 'go', 'here']
    })} {...props} />
  );
}
```

jaortiz: implemented in separate pr :) 
https://github.com/WestpacGEL/GEL/pull/141

components/panel/src/PanelBody.js
```diff
@@ -1,24 +1,14 @@
...
 
 // ==============================
 // Component
 // ==============================
 
-export const PanelBody = ({ appearance, ...props }) => {
-	const { breakpoints, panel } = useTheme();
-	const mq = paint(breakpoints);
+export const PanelBody = props => {
+	const mq = useMediaQuery(breakpoints);
```
simonswiss: We're not passing `breakpoints` anymore here, they're provided in the `useMediaQuery` hook.

```suggestion
      const mq = useMediaQuery();
```

jonnystening: My mistake. Was late 😬

Issue #44, opened 2019-07-10
Component: Table


jonnystening:
## Features
- [x] Bordered prop
- [x] Responsive prop
- [x] Bordered
- [x] Bordered & striped
- [x] Responsive
- [x] Examples
- [ ] Jest tests
- [ ] Cypress tests
- [x] Docs

## To discuss
- [ ] Table component currently only provides the parent wrapping. Row and cell (children) elements are composed with standard html elements. Are there any benefits in abstracting these standard html elements into React components?

- [ ] Responsive table (wrapper) element enabled via prop. Maybe this should be the default? `noResponsive` prop to disable?

- [ ] Table wrapper class `table-responsive` doesn't feel right. Used by Panel component styling; responsive table styling reset when placed inside Panel.

- [ ] `.table-highlight` class doesn't feel very react-esque

- [ ] TableWrapper approach? Should it be in the Table.js file or moved to a styled.js or TableWrapper.js (note: it's not exported)
- [ ] Table can be composed within Panel. A `.table-responsive` className is rendered in case Panel composes a responsive Table (as it requires some style tweaks). Is there another way, without the className?
- [ ] How else can we do the `.table-highlight` on cells/rows?
- [ ] Do we need to check proptypes for a wrapper component that's not exported?

## Feedback from Kate's review
- **Table/examples** -  Wrong styling for _Table caption text_, it's too light 
_JPS: need to discuss, Dom has implemented a solution to handle system fonts not having all weights_
_JPS: This restriction has now been removed_

jaortiz:
Need to further discuss whether we should create a react component for table elements. Currently only using react for the <table> tag and targeting specific cell highlighting via classNames. Not sure if this is the correct way to do this.

jonnystening:
> Need to further discuss whether we should create a react component for table elements. Currently only using react for the
> 
> tag and targeting specific cell highlighting via classNames. Not sure if this is the correct way to do this.

All Table sub-elements (`<tr>`,`<th>`, `<td>` etc) are now in React

PR #143, opened 2019-09-26
Update Table

jonnystening:
## Updates
- Updates styling to use new tokens
- Move style objects inline
- Some tidy up stuff... add missing prop spread on TableWrapper, remove `children` destructuring assignment (saves lines)

## Questions
- TableWrapper approach? Should it be in the Table.js file or moved to a styled.js or TableWrapper.js (note: it's not exported)
- Table can be composed within Panel. A `.table-responsive` className is rendered in case Panel composes a responsive Table (as it requires some style tweaks). Is there another way, without the className?
- How else can we do the `.table-highlight` on cells/rows?
- Do we need to check proptypes for a wrapper component that's not exported?
- See https://github.com/WestpacGEL/GEL/issues/44


components/table/src/Table.js
```diff
@@ -8,157 +8,158 @@ import { jsx, useTheme } from '@westpac/core';
...
-				border: isBordered && 0,
-
-				caption: {
-					padding: table.responsive.caption.padding,
-				},
-				'tbody, tfoot': {
-					'tr:last-child': {
+	return responsive ? (
+		<div
+			className="table-responsive"
```
simonswiss: This will change (we don't want to rely on a className to change styles managed in another component), but it will be easier to merge both the `Table` and `Panel` component changes in `gel-develop`, and open a new PR to _decouple_ them.

simonswiss:
> * TableWrapper approach? Should it be in the Table.js file or moved to a styled.js or TableWrapper.js (note: it's not exported)

Since it's not exported and not intended to use in other places, it's fine to keep it in the same file, makes it a bit easier to grasp in one look. 

If the file becomes too long, it might be an idea to move the Wrapper to its own file.

> * Table can be composed within Panel. A .table-responsive className is rendered in case Panel composes a responsive Table (as it requires some style tweaks). Is there another way, without the className?

Yes! I implemented this in #157 with a `PanelTable` component export, as discussed.

> * How else can we do the .table-highlight on cells/rows?

`<Row highlight />` and `<Cell highlight />` components :)

Checking PropTypes is always good, but when the component is not intended to be used anywhere, it loses a lot of its benefit!

PR #144, opened 2019-09-26
Update Symbol

jonnystening:
## Updates
- Use new tokens
- Update fontSize and margin rem values for examples
- Separate style objects 
- Add new SYMBOL token to power the multi-brand logos
- Combined the LogoSmall, LogoSmallCenter and LogoSmallRight components as LogoSmall, taking an align prop to handle the alignment
- Basic tidy up

## Questions
- I've stored the multi-brand logo's SVG code as strings in a new SYMBOL token. To render this in the multi-brand components (LogoLarge & LogoSmall) I've used `dangerouslySetInnerHTML` on a svg group `<g>` tag, since that's the only way I could work out how to do it. Keen to learn the right way 🤔

- After some thinking (I sometimes do this when running or in the shower)... perhaps the resize feature (`width` and `height` props) is something that Symbol should _not_ be concerned with. We allow sizing of the Icon component via size prop, but that's because we mandate a set of sizes. The sizing of Symbols can be done externally I now reckon.


components/symbol/src/Symbol.js
```diff
@@ -12,36 +12,31 @@ const asArray = val => (Array.isArray(val) ? val : [val]);
...
+
+	const ratio = viewBoxWidth / viewBoxHeight;
+
+	// Common styling
+	const styleCommon = {
+		display: 'inline-block',
+		flexShrink: 0,
+		lineHeight: 1,
+		verticalAlign: 'middle',
+	};
```
jaortiz: inline these styles

PR #145, opened 2019-09-26
Update Badge and Label

jonnystening:
## Updates
- Use new tokens
- Remove children destructure assignment (save lines)
- Rename information appearance as info (note: will be doing this to all components)
- Badge: add print styling

## Todo
- STG brand hero appearance uses `COLORS.text` for text color, other brands use white. This will be a thing throughout a number of components. The default/foreground colour mapping we used to have solved this. How do we proceed?

- Badge styling is not yet adjusted when the Badge is the child of a Button, this feature was in GUI2.0


components/badge/src/Badge.js
```diff
@@ -8,38 +8,38 @@ import { jsx, useTheme } from '@westpac/core';
...
-			borderColor: appearance === 'faint' ? colors.border : colors[appearance].default,
+		'@media print': {
+			color: '#000',
+			backgroundColor: '#fff',
+			border: '1px solid #000',
 		},
 	};
 
-	return (
-		<span css={{ ...style.common, ...style.appearance }} {...props}>
```
jonnystening: What do we reckon here?

Is there any benefit in separating (and then spreading) common styling and styling based on a prop (ie. appearance as in this case)?

components/badge/src/Badge.js
```diff
@@ -8,38 +8,38 @@ import { jsx, useTheme } from '@westpac/core';
...
-		</span>
-	);
+	// Appearance styling
+	const styleAppearance = {
+		color: appearance === 'faint' ? COLORS.muted : '#fff',
+		backgroundColor: COLORS[appearance],
+		borderColor: appearance === 'faint' ? COLORS.border : COLORS[appearance],
+	};
+
+	return <span css={{ ...styleCommon, ...styleAppearance }} {...props} />;
```
jonnystening: What do we reckon here?

Is there any benefit in separating (and then spreading) common styling and styling based on a prop (ie. `appearance` as in this case)?

jaortiz: I would just inline them all. Dont think there is any benefit to separating them here. We were separating them in other components when we could make a mapping to the styles i.e.
```
const appearance  = {
   type1: {},
   type2: {}
}

return <div css={{...appearance[type]}} />
```

components/label/src/Label.js
```diff
@@ -8,77 +8,62 @@ import { jsx, useTheme } from '@westpac/core';
...
-
-			'a&:hover, a&:focus': {
-				cursor: 'pointer',
-				backgroundColor: label.appearance[appearance].hover.backgroundColor,
-				borderColor: label.appearance[appearance].hover.borderColor,
-			},
+		'a&:hover, a&:focus': {
+			cursor: 'pointer',
+			backgroundColor: appearance === 'faint' ? '#fff' : COLORS.tints[`${appearance}50`],
+			borderColor: appearance !== 'faint' && COLORS.tints[`${appearance}50`],
```
jaortiz: inline the styles. Same comment as badge

PR #146, opened 2019-09-26
Update SrSkipLink

jonnystening:
## Updates
- Remove old token values
- Update rem values
- Tidy up


components/accessibility-helpers/src/SrSkipLink.js
```diff
@@ -2,51 +2,48 @@
...
-
-		':focus': {
-			outlineOffset: -link.focus.outlineOffset, //override to be inside
-		},
-	};
-
-	return <a css={common} {...props} />;
-};
+			':focus': {
+				outlineOffset: -2, //override to be inside
```
jonnystening: This offset override depends on the focus outline width setting (defined in Core). 

Should this somehow link to that setting in Core? Or maintain this setting separately and hope they always align?

PR #147, opened 2019-09-26
Update Alert

jonnystening:
## Updates
- Use new tokens
- Update rem values
- Move style objects inline
- Drop isPropname syntax
- Rename information appearance as info

## To fix
- Closable alerts will work after Button and Icon fixes are merged (https://github.com/WestpacGEL/GEL/pull/136) 

## To discuss
- I've implemented a default Alert Icon (`InfoIcon`) for the Alert, since 90% of our Alerts are info alerts. The implementation allows for a different icon to be configured (`icon={HouseIcon}`) or removed completely (`icon={null}`). Is that the right approach? 

- We're going to have to update this to support Icon code-splitting, once that's made available in the icon package.


components/icon/src/Icon.js
```diff
@@ -2,7 +2,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { jsx, useTheme, paint } from '@westpac/core';
+import { jsx, useTheme, useMediaQuery } from '@westpac/core';
```
simonswiss: I have fixed the `Icon.js` file to be able to preview, looking good!

components/alert/src/Alert.js
```diff
@@ -11,93 +11,98 @@ import { CSSTransition } from 'react-transition-group';
...
+		borderTop: '1px solid',
+		borderBottom: '1px solid',
+		borderColor: appearance === 'system' ? COLORS.system : COLORS.tints[`${appearance}50`],
+	};
+
 	return (
 		<CSSTransition in={open} unmountOnExit classNames="anim" timeout={400}>
-			<div css={mq({ ...style.common, ...style.appearance })}>
-				{isClosable && (
+			<div css={mq({ ...styleCommon, ...styleAppearance })}>
```
simonswiss: You can also _merge_ the objects with emotion's array syntax 

```
<div css={mq([ styleCommon, styleAppearance ])}>
```

However the result is a little different, common styles in both common and appearance get both applied, whereas with the spread syntax, new declarations on the same property replace the previous one (which is what you want here I think!)

jonnystening: Yeah I hate seeing CSS overrides within the same class

jonnystening:
OK, let's merge to `gel-develop` and fix any issues afterwards

components/alert/src/Alert.js
jonnystening: Hmm... according to this shouldn't the array syntax merge them?

https://emotion.sh/docs/composition

PR #148, opened 2019-09-27
Update Breadcrumb

florettewest:
Should we check if the children contain the component "Crumb"?

## To discuss
- Will require code-splitting of `ArrowRightIcon` once Icon can support that


components/breadcrumb/src/Breadcrumb.js
```diff
@@ -13,23 +13,23 @@ import { SrOnly } from '@westpac/accessibility-helpers';
...
-
-	const common = {
-		padding: breadcrumb.padding,
-		marginBottom: breadcrumb.marginBottom,
-		fontSize: breadcrumb.fontSize,
-		listStyle: 'none',
-	};
-
 	const childrenWithProps = Children.map(children, (child, index) =>
 		index === Children.count(children) - 1 ? cloneElement(child, { last: 'true' }) : child
```
jonnystening:
```suggestion
		index === Children.count(children) - 1 ? cloneElement(child, { last: true }) : child
```

components/breadcrumb/src/Crumb.js
```diff
@@ -14,38 +14,39 @@ import { SrOnly } from '../../accessibility-helpers/src';
...
+			// <ArrowRightIcon
+			// 	aria-hidden="true"
+			// 	size="small"
+			// 	color={COLORS.primary}
+			// 	css={{
+			// 		marginLeft: '0.1875rem',
+			// 		marginRight: '0.1875rem',
+			// 	}}
+			// />
+			}
```
jonnystening: You can get this working (to test) by fixing the breakpoints desctructure in Icon. There's already a PR for this https://github.com/WestpacGEL/GEL/pull/136

```
...
LAYOUT: { breakpoints },
...
```

components/breadcrumb/src/Crumb.js
```diff
@@ -14,38 +14,39 @@ import { SrOnly } from '../../accessibility-helpers/src';
...
+			css={{
+				display: 'inline-block',
+				position: 'relative',
+				color: COLORS.text,
+
+				a: {
+					color: COLORS.text,
+					textDecoration: 'none',
+
+					'&:focus, &:hover': {
```
jonnystening:
```suggestion
					':focus, :hover': {
```

components/breadcrumb/src/Crumb.js
```diff
@@ -14,38 +14,39 @@ import { SrOnly } from '../../accessibility-helpers/src';
...
 
 	return (
-		<li css={common} {...props}>
+		<li
+			css={{
+				display: 'inline-block',
+				position: 'relative',
+				color: COLORS.text,
+
+				a: {
```
jonnystening:
```suggestion
				'a': {
```

🤷‍♂️

components/breadcrumb/src/Crumb.js
```diff
@@ -14,38 +14,39 @@ import { SrOnly } from '../../accessibility-helpers/src';
  * Breadcrumb: Breadcrumbs are styled navigational links used to indicate a user journey or path. They are a simple, effective and proven method to aid orientation.
  */
 export const Crumb = ({ children, last, ...props }) => {
```
jonnystening: should probably set a proptype for last

`last: PropType.Bool,`

@dominikwilkowski We proptype check everything right? Even these type of props (used internally for styling purposes)

components/breadcrumb/src/Breadcrumb.js
```diff
@@ -13,23 +13,23 @@ import { SrOnly } from '@westpac/accessibility-helpers';
...
-	const common = {
-		padding: breadcrumb.padding,
-		marginBottom: breadcrumb.marginBottom,
-		fontSize: breadcrumb.fontSize,
-		listStyle: 'none',
-	};
-
 	const childrenWithProps = Children.map(children, (child, index) =>
-		index === Children.count(children) - 1 ? cloneElement(child, { last: 'true' }) : child
+		index === Children.count(children) - 1 ? cloneElement(child, { last: true }) : child
```
simonswiss: TL;DR: This is a valid use case for `cloneElement`, despite me recommending not using it.

I keep suggesting to not using `cloneElement`, but when dealing with things you'd do with `Children.count`, as for example here to determine the last child, I think this `cloneElement` is actually the way to go.

If you find yourself in need to know the children index, this is a good use for it.

If you're just passing props that you could grab from the child via Context, I recommend using Context instead (more robust to DOM structure change, doesn't break React's composition model).

jonnystening: Very well explained as always @simonswiss

How about @florettewest’s question. When using cloneElement() should we check the type of the child to ensure props aren’t being passed to random children?

simonswiss: I would say this is a great safe guard. Make sure the correct child is passed, and trow an error when it's not, to prevent accidents from happening.

PR #149, opened 2019-09-27
Update ProgressBar

florettewest:


components/progress-bar/src/ProgressBar.js
```diff
@@ -13,73 +13,80 @@ import { SrOnly } from '@westpac/accessibility-helpers';
  * Progress Bar: A visual indication of progress. Use when loading content or to indicate how far along the user is in a journey.
  */
 export const ProgressBar = ({ value, isSkinny, ...props }) => {
```
jonnystening:
```suggestion
export const ProgressBar = ({ value, skinny, ...props }) => {
```

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,87 @@ import PropTypes from 'prop-types';
...
+					overflow: 'hidden',
+					transition: 'width .6s ease',
+
+					'@media print': {
+						backgroundColor: '#000 !important',
+
+						span: {
+							color: '#fff !important',
+							backgroundColor: '#000 !important',
+						},
```
jonnystening: the print styling for `span` should probably be moved inline into the `<span>` element itself

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,91 @@ import PropTypes from 'prop-types';
...
-				left: '1rem',
-				top: 0,
-				height: '100%',
-				color: '#575f65',
-				fontSize: '1.4rem',
-				fontWeight: 700,
-				zIndex: 1,
-			},
-		},
+	const roundedValue = round(value);
```
simonswiss: I computed the rounded value once here, and use it in 2 place instead of invoking the `round()` function twice with the same argument.

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,91 @@ import PropTypes from 'prop-types';
...
-			borderRadius: progressBar.size[size].height,
-			zIndex: 2,
-			overflow: 'hidden',
-			transition: 'width .6s ease',
-
-			'@media print': {
-				backgroundColor: 'black !important',
+	return (
+		<div
+			css={{
```
simonswiss: Styles moved directly into the css prop, as discussed.

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,91 @@ import PropTypes from 'prop-types';
...
-					color: 'white !important',
-					backgroundColor: 'black !important',
+				'::after': {
+					display: skinny && 'none',
+					content: '"0%"',
+					position: 'absolute',
+					left: '0.625rem',
+					top: 0,
+					height: '100%',
+					color: '#575f65',
```
simonswiss: Should this not be a color pulled from the tokens?

jonnystening: Should be...

```suggestion
					color: COLORS.muted,
```

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,91 @@ import PropTypes from 'prop-types';
...
+					overflow: 'hidden',
+					transition: 'width .6s ease',
+
+					'@media print': {
+						backgroundColor: '#000 !important',
+
+						span: {
+							color: '#fff !important',
+							backgroundColor: '#000 !important',
+						},
```
jonnystening:
```suggestion
```

Can remove these styles, inlined into <span> directly

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,102 @@ import PropTypes from 'prop-types';
...
```
jonnystening: Can remove these styles, inlined into <span> directly

components/progress-bar/src/ProgressBar.js
```diff
@@ -5,90 +5,91 @@ import PropTypes from 'prop-types';
...
+					fontWeight: 700,
+					lineHeight: '1.25rem',
+					color: '#fff',
+					textAlign: 'right',
+					backgroundColor: COLORS.hero,
+					borderRadius: skinny ? '0.625rem' : '1.5rem',
+					zIndex: 2,
+					overflow: 'hidden',
+					transition: 'width .6s ease',
+
```
jonnystening: we concerned about line breaks inside style blocks? Whitespace or no whitespace? (@simonswiss I see you don't like them?)

PR #150, opened 2019-09-27
Update Form and TextInput

jonnystening:
## Updates
- Use new tokens
- Change boolean isPropname syntax,
- Change the way FormContext is created and consumed

Note: Form sub-components (e.g. TextInput) can be used without a Form wrapper so we may not always have FormContext. Removed the throw error from @simonswiss's Context API example in this case.


components/form/src/ErrorMessage.js
```diff
@@ -43,17 +35,17 @@ export const ErrorMessage = ({ message, icon, tag: Tag, ...props }) => {
 	}
 
 	const common = {
-		fontSize: errorMessage.fontSize,
-		margin: errorMessage.margin,
-		color: errorMessage.color,
+		fontSize: '0.875rem',
+		margin: '0 0 0.75rem',
+		color: COLORS.danger,
 		...(isMessages && { listStyle: 'none', paddingLeft: 0 }),
```
jaortiz: can do in one line
```
...(Array.isArray(message) && { listStyle: 'none', paddingLeft: 0 }), 
```

components/form/src/Hint.js
```diff
@@ -1,27 +1,43 @@
...
 	};
 
-	return <Tag css={common} {...props} />;
+	return (
+		<Tag
+			css={{
+				color: COLORS.muted,
+				fontSize: '0.875rem',
+				marginTop: spacingMap[spacing].marginTop,
+				marginBottom: spacingMap[spacing].marinBottom,
```
jaortiz:
```
...spacingMap[spacing]
```

components/form/src/ErrorMessage.js
jonnystening: It's used above though, saves calling isArray twice... thought the variable and comment also made it clearer

components/form/src/Hint.js
jonnystening: whoops also what's a `marinBottom` haha

components/form/src/ErrorMessage.js
jaortiz: Didnt see it being used above, github had it collapsed. Just inline the common styles

PR #152, opened 2019-09-30
WIP: Progress Ropes

jaortiz:


components/progress-rope/examples/00-default.js
```diff
@@ -0,0 +1,68 @@
+import React, { useReducer } from 'react';
+
+import { ProgressRope, ProgressRopeItem } from '../src';
+import { Grid, Cell } from '../../grid/src';
+import { Button } from '../../button/src';
```
simonswiss: These should be imported from 
```
@westpac/grid
@westpac/button
```

Not sure if this works with bolt, but it's definitely how you should do it in the new `yarn workspaces` monorepo.

PR #157, opened 2019-10-02
Make new PanelTable component to handle styling for Table inside Panel

simonswiss:
## Why?

The `Panel` component shouldn't need to deal with styles for a `Table`. They are separate components, separate encapsulated concerns.

This PR brings a new export to the `Table` component: `PanelTable`. It is a component that wraps the `Table` component and handles some style changes based on the `responsive` prop of the Table in question.


components/table/src/PanelTable.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Table } from './';
+
+export const PanelTable = props => (
+	<Table
+		wrappingStyles={
```
jonnystening: wow, that's sweet

components/table/src/Table.js
```diff
@@ -8,157 +8,160 @@ import { jsx, useTheme } from '@westpac/core';
...
-	return isResponsive ? (
-		<div className="table-responsive" css={common}>
-			{children}
-		</div>
+			]}
+			{...props}
+		/>
 	) : (
-		children
+		props.children
```
jonnystening: use `children` like this (`props.children`), instead as `children` via destructure assignment, so then we can simply parse children via props?

`<div {...props} />` instead of `<div {...props}>{children}</div>`

Am I following?

jonnystening: Actually I think that's my code. I'm going cray

components/table/src/Table.js
```diff
@@ -20,15 +20,15 @@ const TableWrapper = ({ bordered, responsive, withinPanel, ...props }) => {
...
 					[xsOnly]: {
 						width: '100%',
-						marginBottom: withinPanel ? 0 : '1.125rem',
+						marginBottom: '1.125rem',
 						overflowY: 'hidden',
 						overflowX: 'auto',
 						// -ms-overflow-style: '-ms-autohiding-scrollbar',
-						border: withinPanel ? 0 : `1px solid ${COLORS.border}`,
+						border: `1px solid ${COLORS.border}`,
 						// -webkit-overflow-scrolling: 'touch',
```
jonnystening: Need this?

components/table/src/Table.js
```diff
@@ -20,15 +20,15 @@ const TableWrapper = ({ bordered, responsive, withinPanel, ...props }) => {
...
-				withinPanel && { border: 0, marginBottom: 0 },
+				extraStyles,
 				{
 					[xsOnly]: {
 						width: '100%',
-						marginBottom: withinPanel ? 0 : '1.125rem',
+						marginBottom: '1.125rem',
 						overflowY: 'hidden',
 						overflowX: 'auto',
 						// -ms-overflow-style: '-ms-autohiding-scrollbar',
```
jonnystening: Need this?

simonswiss: I think by keeping this in - the border will be removed *every time* the Table or PanelTable has the `responsive` prop. 

My understanding is that we only want to remove the border for the `PanelTable`, so I got rid of this ternary, and applied more specific styles in the `PanelTable` component (`!important).

Had to go more specific otherwise the media query wins the specificity battle, since we're not using a classname anymore.

I think it's a correct, intended use of `!important` so it makes it ok :)

Thoughts?

components/table/src/Table.js
simonswiss: see https://github.com/WestpacGEL/GEL/pull/157#discussion_r330805089

jonnystening: referring to commented out `// -webkit-overflow-scrolling: 'touch',`

components/table/src/Table.js
jonnystening: referring to commented out `// -ms-overflow-style: '-ms-autohiding-scrollbar',`

simonswiss: oh! no we don't

components/table/src/Table.js
simonswiss: removed! 👍

PR #158, opened 2019-10-03
ButtonGroup POC

jonnystening:
@simonswiss Is this a better approach to child styling, rather than overriding child button styling via `:first-child`/`:last-child` CSS selectors?

### DIFF
--- a/components/button-group/src/ButtonGroup.js
+++ b/components/button-group/src/ButtonGroup.js
@@ -37,13 +37,15 @@ export const ButtonGroup = ({
 		setChecked(event.target.value);
 	};
 
-	const numChildren = Children.count(children);
 	const childrenWithProps = Children.map(children, (child, index) => {
-		if (index === 0) {
-			return cloneElement(child, { first: true })
+		if (!child.type.isButton) {
+			throw new Error('<ButtonGroup /> only accepts <Button /> as direct children.');
 		}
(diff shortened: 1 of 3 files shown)


jonnystening:
After discussing with @simonswiss, we are best to make use of CSS overrides to provide this first and last child styling, rather than invoking the help of cloneElement() to pass a first/last prop and style based on that.

jonnystening:
...then this came across my desk: https://github.com/WestpacGEL/GEL/pull/148#discussion_r331280867

Hmm 🤔, perhaps the use of cloneElement() in this case (to pass first/last props), is kosher after all?

components/button-group/src/ButtonGroup.js
```diff
@@ -37,6 +37,18 @@ export const ButtonGroup = ({
...
+	const childrenWithProps = Children.map(children, (child, index) => {
+		if (index === 0) {
+			return cloneElement(child, { first: true })
+		}
+		if (index === numChildren - 1) {
+			return cloneElement(child, { last: true })
+		}
+
+		return child;
+	});
```
simonswiss: As per our discussion + my overnight epiphany - this is actually a good use case for `cloneElement`.

Also, multi-line comments FTW!

jonnystening: 🤣

PR #162, opened 2019-10-10
Update/pagination

florettewest:
I am not sure how the pagination items should be presented to the developer... Currently, it is similar to Breadcrumbs: each item is placed in reading order...

### DIFF
--- a/components/pagination/src/Pagination.js
+++ b/components/pagination/src/Pagination.js
@@ -1,0 +1,56 @@
+/** @jsx jsx */
+
+import React, { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useTheme } from '@westpac/core';
+
+// ==============================
+// Utils
+// ==============================
+
+// ==============================
+// Component
+// ==============================
+
+/**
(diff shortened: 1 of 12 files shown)


components/pagination/src/PaginationItem.js
```diff
@@ -0,0 +1,90 @@
...
+						marginLeft: -1,
+						fontSize: '0.875rem',
+						color: active ? '#fff' : COLORS.neutral,
+						zIndex: active && 2,
+						backgroundColor: active && COLORS.hero,
+
+						...(first && {
+							marginLeft: 0,
+							borderTopLeftRadius: '3px',
+							borderBottomLeftRadius: '3px',
```
jonnystening: These px values should be in rems. We made the decision that only border width is in px.

components/pagination/src/Pagination.js
```diff
@@ -0,0 +1,52 @@
...
+		return child;
+	});
+
+	return (
+		<ul
+			css={{
+				display: 'flex',
+				paddingLeft: 0,
+				margin: '1.3125rem 0',
+				borderRadius: '3px',
```
jonnystening: Should be in rems

PR #166, opened 2019-10-11
Update list

jaortiz:


components/list/src/List.js
```diff
@@ -1,25 +1,40 @@
...
-		<ListType css={{ margin: 0, padding: type === 'ordered' ? '0 0 0 2rem' : 0 }}>
-			{childrenWithProps}
-		</ListType>
+		<ListContext.Provider value={listStyle}>
+			<ListType
+				css={{ margin: 0, padding: listStyle.type === 'ordered' ? '0 0 0 1.25rem' : 0 }}
+				{...props}
+			>
+				{children}
+			</ListType>
```
jonnystening: I've been not destructuring children unless needed. Why not just do this...

```suggestion
			<ListType
				css={{ margin: 0, padding: listStyle.type === 'ordered' ? '0 0 0 1.25rem' : 0 }}
				{...props}
			/>
```

components/list/src/List.js
```diff
@@ -1,25 +1,40 @@
...
 // ==============================
 // Component
 // ==============================
-export const List = ({ children, ...props }) => {
-	const childrenWithProps = Children.map(children, child =>
-		isValidElement(child) && child.props.type === 'icon' && child.props.icon
-			? cloneElement(child, { ...props, icon: child.props.icon })
-			: cloneElement(child, props)
-	);
+export const List = ({ appearance, type, spacing, icon, children, ...props }) => {
```
jonnystening: dont need to destructure children if you do the below

PR #170, opened 2019-10-17
Update InputGroup: add default appearance on buttons and change to rems

florettewest:


components/input-group/src/InputGroupButton.js
```diff
@@ -27,8 +27,14 @@ export const InputGroupButton = ({ first, last, ...props }) => {
...
 		/>
 	);
 };
+
+InputGroupButton.defaultProps = {
+	/**
+	 * Button appearance
+	 */
+	appearance: 'hero',
+};
```
jonnystening: Can remove this defaultProp too

PR #171, opened 2019-10-17
Feature/popover

jaortiz:


components/popover/examples/00-default.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Popover } from '../src';
+import { Button } from '@westpac/button';
+const content =
+	'Hello vivamus sagittis lacus vel augue laoreet rutrum faucibus. Lorem ipsum dolor sit amet, consectetur adipisicing elit. Nesciunt laboriosam, mollitia magnam ad magni consequuntur hic et quos optio corrupti praesentium veniam aspernatur minima aperiam ut quas, possimus non architecto. Lorem ipsum dolor sit amet, consectetur adipisicing elit. Aut animi velit in? Suscipit nostrum itaque voluptatibus dolorem qui soluta nobis modi officia incidunt eos dolores atque, unde error delectus officiis.';
+export default () => (
+	<>
+		<Popover title="Popover Title" content={content}>
+			<Button>Click Me</Button>
+		</Popover>
```
jonnystening: Think we need to know more about the ideal composition of a popover. Do we wrap a button toggle or consider as separate elements (as with Modal)

I guess we'll merge and see what TM come back with

components/popover/src/Popover.js
```diff
@@ -0,0 +1,79 @@
...
+	});
+
+	const handleClick = () => {
+		setOpen(!open);
+	};
+
+	return (
+		<>
+			<PopoverPanel open={open} position={position} ref={popoverRef} {...props} />
+			<div css={{ display: 'inline-block' }} ref={triggerRef} onClick={handleClick}>
```
jonnystening: Missing other attributes

`...aria-hidden="true" tabindex="-1" aria-label="Use the ESC key to close"`

components/popover/examples/00-default.js
jaortiz: The way I've developed this it doesnt have a dependency on button or any other component. I just require a child element to act as the reference so I know the position to render the popover.

components/popover/src/Popover.js
jaortiz: Added the label and tab index but I've updated to conditionally render the popover so dont require aria hidden.

PR #178, opened 2019-10-22
Update Switch

florettewest:


components/switch/package.json
```diff
@@ -14,12 +14,10 @@
 		"test:test-dev": "cypress open"
 	},
 	"main": "dist/switch.cjs.js",
-	"module": "dist/switch.esm.js",
```
dominikwilkowski: Let's not remove the module path here. Still need it :)

PR #179, opened 2019-10-23
Feature/progress ropes

jaortiz:


jaortiz:
@dominikwilkowski the changes in the other files were from running prettier so I could pass CI. I have reverted the commit so that the diff should only has the progress-rope files now.

components/progress-rope/examples/00-default.js
```diff
@@ -0,0 +1,41 @@
+/** @jsx jsx */
+
+import { GEL, jsx } from '@westpac/core';
+import { Grid, Cell } from '@westpac/grid';
+import { Button } from '@westpac/Button';
+import { ProgressRope, Item } from '../src';
```
dominikwilkowski: Can you use the absolute import here please. It's better for readability and docs
```suggestion
import { ProgressRope, Item } from '@westpac/progress-rope';
```

components/progress-rope/examples/10-group.js
```diff
@@ -0,0 +1,77 @@
+/** @jsx jsx */
+
+import { GEL, jsx } from '@westpac/core';
+import { Grid, Cell } from '@westpac/grid';
+import { Button } from '@westpac/Button';
+import { ProgressRope, Group, Item } from '../src';
```
dominikwilkowski:
```suggestion
import { ProgressRope, Item } from '@westpac/progress-rope';
```

components/progress-rope/src/ProgressRope.js
```diff
@@ -0,0 +1,103 @@
...
+// ==============================
+// Component
+// ==============================
+export const ProgressRope = ({ current, children, ...props }) => {
+	const [progress, setProgress] = useState({
+		currStep: current,
+		currGroup: 0,
+		openGroup: 0,
+		...createRopeGraph(children),
+	});
```
dominikwilkowski: Can we do this in separate states? Maybe use `useReducer`?

components/progress-rope/src/ProgressRope.js
```diff
@@ -0,0 +1,103 @@
...
+
+	const handleClick = index => {
+		setProgress({ ...progress, openGroup: index !== progress.openGroup ? index : null });
```
dominikwilkowski: This would come up as stale state here. Easier to separate them into each states or `useReducer`

components/progress-rope/src/ProgressRopeGroup.js
```diff
@@ -0,0 +1,79 @@
+/** @jsx jsx */
+
+import { Children, cloneElement } from 'react';
+import PropTypes from 'prop-types';
+import { jsx, useBrand } from '@westpac/core';
+import { useProgressRopeContext } from './ProgressRope';
+
+export const ProgressRopeGroup = ({ index, label, children, ...props }) => {
+	const { openGroup, ropeGraph, handleClick } = useProgressRopeContext();
```
dominikwilkowski: We may not need context here?

jaortiz: Why not? I'm using those values calculated in the parent ProgressRope component

jaortiz:
@dominikwilkowski I've updated to useReducer. I'm calling dispatch multiple times in a row. Is there a nicer way to do this or is it fine this way?

components/progress-rope/src/ProgressRopeGroup.js
dominikwilkowski: Can't just use props? Too many?

jaortiz: Im a bit confused here. I would have to clone the child component to pass props to it? I thought we were moving away from that pattern. 
```
<ProgressRope current={indexOfStep}>
   <ProgressRopeGroup>
      <ProgressRopeItem>Step</ProgressRopeItem>
   </ProgressRopeGroup>
</ProgressRope>
```
This is how I've structured the usage of the component. The values I'm grabbing via context are calculated in the ProgressRope component based on the index of the step that is passed in via props.

dominikwilkowski: I'll have a proper look tomorrow. No blockers from me. Happy to merge for now

PR #180, opened 2019-10-23
Spike: button group radio behaviour

jossmac:
Most of the time(~90%) a consumer will be using a button group outside of a form, or at least not with the intent to submit the value with the form. This PR aims to simplify the implementation and reduce the HTML shipped to the client for each instance.

- [x] simplify the `ButtonGroup` implementation and expose from `@westpac/button`
- [ ] ~expose `RadioGroup` from `@westpac/form-check` with the ability to look like a button group~
- [x] remove the `@westpac/button-group` package
- [x] support the `data` pattern

Supports:
- uncontrolled and controlled behaviours via `value | onChange` properties
- `defaultValue` for uncontrolled
- value of type “index” `integer`
- value of type “key” `string`, expects `value` on child buttons
- form support via `name` property and hidden field
- preserved `onClick` handlers for all child buttons

Related to #41


components/button/src/ButtonGroup.js
```diff
@@ -0,0 +1,125 @@
...
+		look,
+		name,
+		onChange,
+		value: controlledValue,
+		size,
+		...rest
+	} = props;
+	const [value, setValue] = useState(defaultValue);
+
+	devWarning(children && data, 'ButtonGroup accepts either `children` or `data`, not both.');
```
jossmac:
```suggestion
	devWarning(children && data, 'ButtonGroup accepts either `children` or `data`, not both.');
	devWarning(!children && !data, 'ButtonGroup expects either `children` or `data`.');
```

PR #184, opened 2019-10-29
Update Popover

jaortiz:


components/popover/src/Popover.js
```diff
@@ -43,7 +49,6 @@ export const Popover = ({ open: isOpen, children, ...props }) => {
 		}
 	};
 
-	// not sure if this is correct
 	useEffect(() => {
```
jossmac: This looks like a great opportunity for a custom hook like `useOutsideHook` we could export from core.

```jsx
const handler = () => {}
const ref = useOutsideClick(handler)

return <div ref={ref} />
```

jaortiz: Good idea, ill add it in a separate pr.

components/popover/src/Popover.js
```diff
@@ -1,19 +1,25 @@
 /** @jsx jsx */
 
-import React, { useState, useEffect, useRef } from 'react';
+import { useState, useEffect, useRef, Fragment } from 'react';
 import PropTypes from 'prop-types';
-import { jsx } from '@westpac/core';
+import { jsx, useBrand, merge } from '@westpac/core';
+import pkg from '../package.json';
```
jossmac: I think this is making the build cranky

jaortiz: I just need to run prettier on the package.json. Always happens when I do a yarn add, prettier doesnt like the way yarn formats adding a package. Ill make sure it passes build once I'm finished updating this component with overwrite tokens :smile:

PR #185, opened 2019-10-29
Focus example in core

jonnystening:
Typo in web transformer

`weight` --> `font-weight`
`style` --> `font-style`

Update type token and focus examples


jonnystening:
Ok no probs

Have added a `<Body />` wrapped button in the example. We should probably cater for link buttons placed within text blocks. How do we fix the button styling (ie. color and text-decoration affected by link styling)?

PR #187, opened 2019-11-04
Update tabcordion

jaortiz:


components/tabcordion/examples/50-local-tokens.js
```diff
@@ -0,0 +1,116 @@
+/** @jsx jsx */
+
+import { forwardRef } from 'react';
+import { GEL, jsx } from '@westpac/core';
+import { Tab, Tabcordion } from '@westpac/tabcordion';
+import { data } from './_data';
+
+const Panel = forwardRef(({ isLast, isSelected, ...props }, ref) => {
+	const common = {
```
dominikwilkowski: Not to be silly but can't we just put the thing in there? What's the point of the `common` var here?

components/tabcordion/examples/50-local-tokens.js
```diff
@@ -0,0 +1,116 @@
...
+		backgroundColor: 'black',
+		color: 'white',
+		padding: 16,
+	};
+	return <div ref={ref} css={common} {...props} />;
+});
+
+const TabItem = ({ isSelected, isJustified, isLast, ...props }) => {
+	const common = {
```
dominikwilkowski: Same as above... This is harder for me to connect

components/tabcordion/examples/50-local-tokens.js
```diff
@@ -0,0 +1,116 @@
...
+		fontSize: '1rem',
+		justifyContent: 'space-between',
+		outline: 0,
+		padding: '0.75rem 1.125rem',
+		position: 'relative',
+		textAlign: 'left',
+		width: '100%',
+	};
+
+	const lastStyles =
```
dominikwilkowski: Same ☝️

components/tabcordion/examples/50-local-tokens.js
```diff
@@ -0,0 +1,116 @@
...
+		Panel,
+		TabItem,
+		AccordionLabel,
+	};
+
+	return (
+		<GEL brand={overridesWithTokens}>
+			<h3>Responsive</h3>
+			<Tabcordion>
+				{data.map(t => (
```
dominikwilkowski: I'm actively removing automated layouting from example files so it's easier to read. Let's repeat things so people can the the actual accordion API

components/tabcordion/src/Tabcordion.js
```diff
@@ -1,9 +1,10 @@
 import React, { Children, useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
+import { useBrand, merge } from '@westpac/core';
 import { useContainerQuery } from '@westpac/hooks';
-
-import { TabItem, TabRow } from './styled';
+import { TabItem, TabRow, AccordionLabel, Panel } from './styled';
```
dominikwilkowski: We need to talk about `styled.js`.

components/tabcordion/src/Tabcordion.js
```diff
@@ -1,9 +1,10 @@
...
 import { Tab } from './Tab';
+import pkg from '../package.json';
 
 let instanceId = 0;
```
dominikwilkowski: We need to make sure this works when the component is run multiple times on the same page even in different react wrappers (multiple react apps)

components/tabcordion/src/Tabcordion.js
```diff
@@ -87,13 +98,13 @@ export const Tabcordion = props => {
 	const tabCount = Children.count(props.children);
 
 	return (
-		<div ref={containerRef}>
+		<div ref={containerRef} css={{ ...overrides.CSS }}>
```
dominikwilkowski: Nice but we said we would lowercase `css` when it stands alone: https://github.com/WestpacGEL/GEL/tree/develop#local-tokens-naming-convention-overwrites

components/tabcordion/src/styled.js
```diff
@@ -17,78 +17,62 @@ export const TabRow = forwardRef((props, ref) => (
...
+				textAlign: 'left',
+				textDecoration: 'none',
+				transition: 'background .3s ease',
+				width: '100%',
+				cursor: 'pointer',
+
+				'&:last-child': {
+					marginRight: 0,
+				},
+				...styles[appearance],
```
dominikwilkowski: `appearance` should be `look` now: https://github.com/WestpacGEL/GEL/tree/develop#props-api-vocabulary

components/tabcordion/examples/50-local-tokens.js
jaortiz: Yea sorry I forgot to move it into the element itself. I find it easier to move out the styles when I'm working on a component and then move them back when im done.

components/tabcordion/src/Tabcordion.js
jaortiz: Are we moving all styled components into the component file that requires them?

components/tabcordion/src/Tabcordion.js
jaortiz: How do we want to handle unique id's for components? You said before you didnt like generating a random unique id (was using shortid package in other components to do this.)

components/tabcordion/src/Tabcordion.js
dominikwilkowski: We should chat about this. I personally like this more :)

components/tabcordion/src/Tabcordion.js
dominikwilkowski: We need to add the unique ID to the `window` object directly.

jaortiz:
I've fixed most of the PR comments. Only ones I haven't done are the ones that require further discussion:
- styled.js
- unique id in the window object

components/tabcordion/src/Tab.js
```diff
@@ -1,50 +1,61 @@
-import React, { forwardRef } from 'react';
+/** @jsx jsx */
+
+import React, { useState, useEffect, forwardRef } from 'react';
```
dominikwilkowski: `useEffect` is declared but never used

components/tabcordion/src/Tab.js
```diff
@@ -1,50 +1,61 @@
...
+import { jsx, useBrand, merge, wrapHandlers } from '@westpac/core';
```
dominikwilkowski: `wrapHandlers` is declared but never used

PR #188, opened 2019-11-04
add useOutsideClick

jaortiz:


dominikwilkowski:
I think this should go into the hooks package no?

jaortiz:
Moved :bowtie:

PR #189, opened 2019-11-04
Update symbol

jaortiz:


jaortiz:
I made some changes to buildExports.
- Made more generic so it doesn't just use icon naming
- Added a check to see if icon/symbol folder exists and create folder + package.json if need be

components/symbol/examples/40-local-tokens.js
```diff
@@ -0,0 +1,55 @@
...
+import {
+	MastercardAcceptedSymbol,
+	BPayLandSymbol,
+	AppleStoreSymbol,
+	WBCLogo,
+	STGLogo,
+} from '@westpac/symbol';
+import { Cell, Grid, Name } from './_utils';
+
+const Wrapper = ({ children, ...props }) => (
```
dominikwilkowski:
```suggestion
const Wrapper = ({ children, symbol, ...props }) => (
```

components/symbol/examples/40-local-tokens.js
```diff
@@ -0,0 +1,55 @@
...
+	<div
+		css={{
+			border: '2px solid red',
+			padding: '24px',
+			height: '100%',
+		}}
+	>
+		{children}
```
dominikwilkowski:
```suggestion
		{children} - {symbol}
```

dominikwilkowski: Let's show in the examples that you have access to this `symbol` prop now in the `Wrapper`.

components/symbol/src/Symbol.js
```diff
@@ -49,8 +50,16 @@ export const Symbol = ({
 	children,
 	...props
 }) => {
+	const { [pkg.name]: overridesWithTokens } = useBrand();
+
+	const overrides = {
+		Wrapper: SymbolWrapper,
+		svgAttributes: {},
```
dominikwilkowski: We also need to spread css for each component. I forgot that in `Icon` as well.

```suggestion
		svgAttributes: {},
		css {},
```

components/symbol/src/Symbol.js
```diff
@@ -64,10 +73,11 @@ export const Symbol = ({
 				role="img"
 				focusable="false"
 				style={{ width: '100%', height: '100%' }}
+				{...overrides.svgAttributes}
 			>
```
dominikwilkowski:
```suggestion
			css={overrides.css}
			>
```

helpers/buildExports.js
```diff
@@ -6,28 +6,31 @@ const fs = require('fs');
 /**
  * Getting all files from a directory
  *
- * @param  {string} iconPath - The path to the folder
+ * @param  {string} svgPath - The path to the folder containing all component svgs
+ *
```
dominikwilkowski:
```suggestion
```

helpers/buildExports.js
```diff
@@ -6,28 +6,31 @@ const fs = require('fs');
...
+ * @param	{string} component - The component name i.e. icons, symbols
```
dominikwilkowski:
```suggestion
 * @param  {string} component - The component name i.e. icons, symbols
```

helpers/buildExports.js
```diff
@@ -6,28 +6,31 @@ const fs = require('fs');
...
  *
  * @return {array}           - an array of all files inside a folder with the ".js" extension stripped
```
dominikwilkowski:
```suggestion
 * @return {array}                       - an array of all files inside a folder with the ".js" extension stripped
```

helpers/buildExports.js
```diff
@@ -6,28 +6,31 @@ const fs = require('fs');
 /**
  * Getting all files from a directory
  *
- * @param  {string} iconPath - The path to the folder
+ * @param  {string} svgPath - The path to the folder containing all component svgs
```
dominikwilkowski:
```suggestion
 * @param  {string} svgPath      - The path to the folder containing all component svgs
```

helpers/buildExports.js
```diff
@@ -6,28 +6,31 @@ const fs = require('fs');
...
+	console.info(chalk.green(`✅ Got all ${component} successfully`));
+	return svgs;
 }
 
 /**
- * Insert an array of icons into a js export file
+ * Insert an array of component svgs into a js export file
  *
- * @param  {array}  icons     - An array of icons
+ * @param  {array}  svgs     - An array of component svgs
```
dominikwilkowski:
```suggestion
 * @param  {array}  svgs      - An array of component svgs
```

helpers/buildExports.js
```diff
@@ -59,38 +62,72 @@ function writePkg(pkgPath, content) {
 }
 
 /**
- * Insert an array of icons into the preconstruct entrypoint array of a package.json
+ * Insert an array of component svgs into the preconstruct entrypoint array of a package.json
  *
- * @param  {array}  icons   - An array of icons
+ * @param  {array}  svgs   - An array of component svgs
```
dominikwilkowski:
```suggestion
 * @param  {array}  svgs    - An array of component svgs
```

helpers/buildExports.js
```diff
@@ -59,38 +62,72 @@ function writePkg(pkgPath, content) {
...
 
 	console.info(chalk.green('✅ package.json file written successfully'));
 }
 
 /**
- * Fix the source key inside each icons package.json file
+ * Fix the source key inside each svgs package.json file
  *
- * @param  {array} icons - An array of all icons
+ * @param  {array} svgs - An array of all component svgs
```
dominikwilkowski:
```suggestion
 * @param  {array} svgs  - An array of all component svgs
```

helpers/buildExports.js
```diff
@@ -59,38 +62,72 @@ function writePkg(pkgPath, content) {
...
-			pkg.preconstruct = {};
+function fixSource(svgs, component) {
+	svgs.map(svg => {
+		// check if package exists
+		const folderPath = path.normalize(`${process.cwd()}/${svg}`);
+		if (!fs.existsSync(folderPath)) {
+			// have a log about creating folder here
+			// need to have error handlers here
+			try {
+				fs.mkdirSync(folderPath);
```
dominikwilkowski: All of this is done inside `preconstruct` and wasn't needed here :)

helpers/buildExports.js
```diff
@@ -59,38 +62,72 @@ function writePkg(pkgPath, content) {
...
+			} catch (error) {
+				console.error(chalk.red(`❌ ${error}`));
+				process.exit(1);
+			}
+
+			const package = {
+				main: `dist/${component.slice(0, -1)}.cjs.js`,
```
dominikwilkowski: I think we need to be careful with duplicating this in `preconstruct` and here... `preconstruct` might change this in the future.

helpers/buildExports.js
jaortiz: It wasn't creating folders for me before thats why I added it in. Ill check with new version

dominikwilkowski: Yeah that's why I said we need to run it twice. It does create it even in the old version but you need to run it first.

jaortiz: I removed this part, removed one of the symbol folders and ran the build script and it errors out like before
```
Error: Cannot find module '/Users/Jeremy/Dev/gel-symbol/components/symbol/AmericanExpressSymbol/package.json'
```

dominikwilkowski: Did yoiu run `preconstruct`?

jaortiz: ahhh I was always running preconstruct dev not build. All good I'll remove it.

dominikwilkowski: sorry that's my fault for not communicating properly... 😁

PR #192, opened 2019-11-07
add overrides and fix css in progress-bar

jaortiz:


components/progress-bar/src/ProgressBar.js
```diff
@@ -116,10 +131,10 @@ ProgressBar.propTypes = {
 	/**
 	 * Enable skinny mode
 	 */
-	skinny: PropTypes.bool,
+	look: PropTypes.oneOf(['default', 'skinny']),
```
dominikwilkowski: If a Prop has a default value it should be marked as `.isRequired`
```suggestion
	look: PropTypes.oneOf(['default', 'skinny']).isRequired,
```

PR #195, opened 2019-11-19
Add overrides to text-input

jaortiz:


jaortiz:
I've made an attempt make these components controlled. Not sure what pattern we want to use for doing this. More than happy for discussion/input on this.

components/text-input/examples/30-fieldset.js
```diff
@@ -12,31 +12,31 @@ function Example({ brand }) {
 			<h2>Default</h2>
 			<fieldset>
 				<legend>Legend</legend>
-				<TextInput />
+				<Text />
 				<br />
-				<TextInput tag="select">
+				<Select>
 					{options.map((v, i) => (
```
dominikwilkowski: Inside examples I would avoid loops that make it hard to read the code. We will likely use these examples on the website soon too.

dominikwilkowski: I'll fixed it up quickly

PR #197, opened 2019-11-26
Update/table

dominikwilkowski:


dominikwilkowski:
I merge this for now but we need to still add more overrides for each component and data-driven capabilities

PR #208, opened 2019-12-04
Further fixes

jonnystening:
**Panel**
- Fixes STG panel header text color

**Form Check**
- Use PACKS for focus
- Fix tick positioning (missing boxSizing reset)

**Symbol**
- Remove WindowsStoreSymbol (now called Microsoft Store, already provided)

**Body**
- New tag prop

**Badge**
- Rename appearance example, now look

**Tooltip**
- New tag prop
- Use title attribute (when abbr tag)
- Remove border reset
- Add Body wrap in example page

**TYPE token**
- Allow all body font weights

**Button**
- rename label prop (assistiveText)
- use aria-label for alternative text, rather than VisuallyHidden (checked with Intopia 👍)

**Icon**
- rename label prop (assistiveText)

**Switch**
- disable focus on click
- fix sizing and positioning
- re-style toggle text positioning
- use aria-label for alternative text, rather than VisuallyHidden (checked with Intopia 👍)

**Breadcrumb**
- typo


components/body/src/Body.js
```diff
@@ -11,7 +11,7 @@ import pkg from '../package.json';
 /**
  * Body: Body component in charge of body text
  */
-export const Body = ({ children, ...props }) => {
+export const Body = ({ tag: Tag, children, ...props }) => {
```
dominikwilkowski: Can you make sure that we include these things in the examples right when we build them?
So in the `00-typography.js` file I'd include two examples: one where you change the `div` to something like `section` and another that takes a component and passes that into the `tag` prop.

components/panel/src/PanelHeader.js
```diff
@@ -21,7 +21,7 @@ export const PanelHeader = props => {
 
 	const lookMap = {
 		hero: {
-			color: '#fff',
+			color: BRAND === 'STG' ? COLORS.text : '#fff',
```
dominikwilkowski: That's the sort of thing that will now go away :)
So that work was for nothing but good to take note of it to not forget it in the brand STG package

components/tooltip/src/Tooltip.js
```diff
@@ -10,12 +10,12 @@ import { TooltipBubble } from './TooltipBubble';
...
 // Overwrite component
 // ==============================
-const Wrapper = forwardRef((props, ref) => <span ref={ref} {...props} />);
+const Wrapper = forwardRef(({ tag: Tag, ...props }, ref) => <Tag ref={ref} {...props} />);
 
 // ==============================
 // Component
 // ==============================
-export const Tooltip = ({ text, ...props }) => {
+export const Tooltip = ({ tag, text, ...props }) => {
```
dominikwilkowski: Same as above we need a test case for this change.

components/body/src/Body.js
jonnystening: Sure 👍

components/panel/src/PanelHeader.js
jonnystening: Did that before you built the new approach

components/body/src/Body.js
jonnystening: With the latter, we talking...

```
const BodyTag = props => <span {...props} />;

...

<Body tag={BodyTag}>
   <p>Body with a component tag</p>
</Body>
```

components/tooltip/src/Tooltip.js
jonnystening: Have added a test case https://github.com/WestpacGEL/GEL/pull/208/commits/9fcc2d7873c5243834a144aa767c1fd4d563d480, but it's throwing a wobbly. Mind looking in to that @jaortiz

PR #214, opened 2019-12-19
WIP: Fix accessiblity in tooltip and popover

jaortiz:


components/tooltip/src/Tooltip.js
```diff
@@ -97,7 +99,7 @@ Tooltip.propTypes = {
 	/**
 	 * Tooltip tag
 	 */
-	tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
+	// tag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
```
jonnystening: Wanna remove this?

jonnystening:
Sorry @jaortiz ... it also needs to programmatically set focus on the bubble when it opens

jaortiz:
@jonnystening  Do we want to focus trap the tooltip/popover? (like in modal?)

jonnystening:
Only need to focus trap when we’re isolating the tabbable items on a UI layer. Usually only when a dark overlay background Is shown... modal and off-canvas navs

jonnystening:
> Only need to focus trap when we’re isolating the tabbable items on a UI layer. Usually only when a dark overlay background Is shown... modal and off-canvas navs

Sorry, ignore this. Yes we need to focus trap. Just in a kind of different way to how its done in Modal

PR #218, opened 2020-01-15
Accessibility review fixes

jaortiz:


components/pagination/examples/00-default.js
```diff
@@ -46,13 +46,11 @@ function Example({ brand }) {
 				back={{
 					visible: true,
 					text: 'Go back',
-					assistiveText: page => `Go to previous page which is ${page}`,
```
dominikwilkowski: Can we not remove the examples? Rather let's make them better or people don't know about the API?!

jaortiz: In the intopia review notes, they said that the text used for the aria-label in the back/next buttons should be matching with the actual button text. So I changed the code to reflect this and therefore the assistiveText option isnt being used anymore.

dominikwilkowski: oh... I see.. we removed it completely? ok cool ... I guess

components/pagination/examples/00-default.js
```diff
@@ -47,11 +47,12 @@ function Example({ brand }) {
 					visible: true,
 					text: 'Go back',
 					onClick: (event, page) => console.log(`Go to ${page}`, event),
+					assistiveText: page => `Go to previous page which is ${page}`,
 				}}
 				next={{
 					visible: true,
 					text: 'Go forth',
-					onClick: (event, page) => console.log(`Go to ${page}`, event),
```
dominikwilkowski: Why remove the `onClick`?

jaortiz: That was an accident. Added it back

dominikwilkowski: easy!

PR #229, opened 2020-03-05
Refactor spread

jaortiz:
@dominikwilkowski After having used the new spread pattern for a bit we've found ourselves having some issues with it. We find ourselves missing props in the destructure or accidentally destructuring something that we actually want spread on a component. 

Im proposing the following refactor:
```js
// Destructure after override reconciliation to remove overrides.ComponentName.component convention
<AccordionLabel 
// Only explicitly pass the props the component requires to work and that we want to be spread
  id={tabId}
  onClick={handleAccordionClick}
  aria-controls={panelId}
  aria-expanded={!hidden}

  // Instead of manually re-passing all props, just pass the state object so we only have to destructure the one prop in overrides
  state={state}

  // attributes and css as as usual with updated name from destructure
  {...accordionLabelAttributes(state)}
  css={accordionLabelStyles(state)}
/>
```
We are also finding that we have variable name conflicts when passing to mutated/cloned children so I'm also proposing that we don't pass the state object to them and that we only pass state to the overrides contained in the component we are in. 

I've refactored a couple components as an example and to test that everything still works. Keen to hear your thoughts.


dominikwilkowski:
Thanks @jaortiz 
I'll have a look soon. Already have thoughts but need to make sure I understand your problem properly first :)

components/body/src/Body.js
```diff
@@ -3,29 +3,20 @@
...
 import PropTypes from 'prop-types';
 
-import { Body as BodyWrapper, bodyStyles } from './overrides/body';
+import { defaultBody } from './overrides/body';
 import pkg from '../package.json';
 
 // ==============================
 // Component
-//
-// Body component in charge of body text
```
dominikwilkowski: yeah who needs this anyway :)

components/breadcrumb/examples/10-overrides.js
```diff
@@ -56,14 +56,25 @@ function Example({ brand }) {
 			>
 				<Crumb href="#/" text="Home" />
 				<Crumb href="#/personal-banking/" text="Personal" />
-				<Crumb href="#/credit-cards/" text="Credit cards" />
+				<Crumb
+					overrides={{
+						CrumbRoot: {
```
dominikwilkowski: Maybe we should just call it `Root`?
I'm still not sure about this `[component]Root` thing...

components/breadcrumb/src/Breadcrumb.js
```diff
@@ -1,18 +1,32 @@
...
-// Component
+// Context and Consumer Hook
 // ==============================
+const BreadcrumbContext = createContext();
+
+export const useBreadcrumbContext = () => {
+	const context = useContext(BreadcrumbContext);
+
+	if (!context) {
+		throw new Error('Breadcrumb sub-components should be wrapped in <Breadcrumb>.');
```
dominikwilkowski:
```suggestion
		throw new Error('<Crumb/> components should be wrapped in <Breadcrumb>.');
```

components/breadcrumb/src/Crumb.js
```diff
@@ -2,102 +2,61 @@
 
 import { jsx, useBrand, overrideReconciler } from '@westpac/core';
 import PropTypes from 'prop-types';
-import React from 'react';
 
-import { Crumb as CrumbWrapper, crumbStyles } from './overrides/crumb';
-import { Link, linkStyles } from './overrides/link';
-import { Icon, iconStyles } from './overrides/icon';
+import { useBreadcrumbContext } from './Breadcrumb';
```
dominikwilkowski: Can we keep the overrides visually separate from the dependencies (in this case the context)?

components/breadcrumb/src/Crumb.js
```diff
@@ -2,102 +2,61 @@
...
-	href,
-	text,
-	assistiveText,
-	onClick,
-	overrides: componentOverrides,
-	...rest
-}) => {
+export const Crumb = ({ current, href, text, onClick, overrides, ...rest }) => {
 	const {
-		COLORS,
```
dominikwilkowski: Good catch!

components/breadcrumb/src/overrides/icon.js
```diff
@@ -1,15 +1,31 @@
...
 });
+
+const iconAttributes = () => {
+	const { COLORS } = useBrand();
+
+	return {
+		'aria-hidden': 'true',
+		assistiveText: null, //remove icon's `aria-label`
+		size: 'small',
+		color: COLORS.primary,
```
dominikwilkowski: Like we discussed that's a prop, not an attribute
```suggestion
```

components/alert/src/Alert.js
```diff
@@ -81,12 +60,13 @@ export const Alert = ({
...
 	};
 
-	const overrides = overrideReconciler(
-		defaultOverrides,
-		tokenOverrides,
-		brandOverrides,
-		componentOverrides
-	);
+	const {
+		AlertRoot: { component: AlertRoot, styles: alertRootStyles, attributes: alertRootAttributes },
```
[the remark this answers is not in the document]
jonnystening: Yeah we end up with unprettier 😢

components/alert/src/overrides/closeBtn.js
```diff
@@ -40,3 +31,11 @@ export const closeBtnStyles = (_, {}) => {
 		},
 	})[0];
 };
+
+const closeBtnAttributes = () => ({ 'aria-label': 'Close' });
```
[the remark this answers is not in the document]
jonnystening: fix up / break 😬

components/alert/src/overrides/icon.js
```diff
@@ -29,10 +17,18 @@ export const Icon = ({
 		return null;
 	}
 
-	return <Tag size={size} color={color} assistiveText={null} aria-hidden={true} {...rest} />;
+	return (
+		<Tag
+			// size={['small', 'medium']}
+			// color="inherit"
```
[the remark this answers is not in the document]
jonnystening: We'll tidy these up as we go through rolling it out

components/breadcrumb/examples/10-overrides.js
jaortiz: I would prefer Root but a component can have multiple Root components and overrides wouldnt work. Happy to hear suggesstions on naming, this was the nicest one I could think of at the time.

components/alert/src/overrides/icon.js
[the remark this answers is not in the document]
jaortiz: Yea. I left that there before because I was unsure where they belong but @dominikwilkowski has already cleared that up for me :grin:

components/breadcrumb/examples/10-overrides.js
dominikwilkowski: Can they? Where? Maybe in a11y? But otherwise each component has a wrapping element no?

jaortiz: In breadcrumb for example we export <Breadcrumb/> and <Crumb/> if both have the wrapping element named as <Root/> how would we do the overrides for them since we do overrides based on the package name i.e.
```
overridesWithTokens['@westpac/breadcrumb'] = {
  Root: {}, 
  Root: {}, 
}
```

dominikwilkowski: Fair point! Let's leave it as you had it

jaortiz: @jonny found out that we can just use the component name as the root component (i.e. `<Breadcrumb/>` instead of `<BreadcrumbRoot/>`). Would there be any potential issues with declaring and using a variable with the same name as the function that contains it @dominikwilkowski ?

PR #247, opened 2020-03-18
Text element style

maryam-mv:


website/src/components/pages/single-component/blocks-hub.js
```diff
@@ -12,6 +12,12 @@ const Bold = props => <strong css={{ fontWeight: 'bold' }} {...props} />;
 const Italic = props => <em css={{ fontStyle: 'italic' }} {...props} />;
 const Strike = props => <span css={{ textDecoration: 'strike-through' }} {...props} />;
 const Under = props => <span css={{ textDecoration: 'underline' }} {...props} />;
+const leftIndent = {
+	paddingLeft: '5rem',
```
MadeByMike: Can we use the spacing helpers?

website/src/components/pages/single-component/blocks-hub.js
```diff
@@ -12,6 +12,12 @@ const Bold = props => <strong css={{ fontWeight: 'bold' }} {...props} />;
...
+	'@media (max-width: 568px)': {
+		paddingLeft: '0rem',
```
MadeByMike:
```suggestion
		paddingLeft: 0,
```

website/src/components/pages/single-component/blocks-hub.js
```diff
@@ -12,6 +12,12 @@ const Bold = props => <strong css={{ fontWeight: 'bold' }} {...props} />;
 const Italic = props => <em css={{ fontStyle: 'italic' }} {...props} />;
 const Strike = props => <span css={{ textDecoration: 'strike-through' }} {...props} />;
 const Under = props => <span css={{ textDecoration: 'underline' }} {...props} />;
+const leftIndent = {
+	paddingLeft: '5rem',
+	'@media (max-width: 568px)': {
```
MadeByMike: Are breakpoints in the design tokens?

website/src/components/pages/single-component/blocks-hub.js
dominikwilkowski: hah I missed those. nice catch

website/src/components/pages/single-component/blocks-hub.js
dominikwilkowski: Yes but that doesn't mean you can add breakpoints anywhere where it makes sense

PR #252, opened 2020-03-19
Add new dynamic blocks

MadeByMike:
- Add image block
- Add image filter block
- Add screen reader text block

### DIFF
--- a/website/dynamic-blocks/image.js
+++ b/website/dynamic-blocks/image.js
@@ -1,0 +1,99 @@
+/** @jsx jsx */
+import React, { useState, useEffect } from 'react'; // Needed for within Keystone
+import { FieldContainer, FieldLabel, FieldInput } from '@arch-ui/fields';
+import { inputStyles } from '@arch-ui/input';
+import { jsx, useBrand } from '@westpac/core';
+import { useMutation } from '@apollo/react-hooks';
+
+import gql from 'graphql-tag';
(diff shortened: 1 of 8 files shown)


website/dynamic-blocks/image.js
```diff
@@ -0,0 +1,99 @@
...
+					<FieldLabel htmlFor={'image-caption'} field={{ label: 'Caption', config: {} }} />
+					<FieldInput>
+						<input
+							css={inputStyles}
+							type="text"
+							id="image-caption"
+							value={caption}
+							onChange={async e => {
+								setCaption(e.target.value);
+							}}
```
dominikwilkowski: that's not an async function is it?

```suggestion
							onChange={e => {
								setCaption(e.target.value);
							}}
```

website/src/components/a11y-filter.js
```diff
@@ -0,0 +1,126 @@
+/** @jsx jsx */
+import React, { useState } from 'react';
+import { jsx, useBrand } from '@westpac/core';
+import Select from 'react-select';
+
+const SVGFilters = () => (
+	<svg
+		aria-hidden="true"
+		className="svg-filter sronly"
```
dominikwilkowski: Why `sronly` or `svg-filter` for that matter? I think that's copy pasta
```suggestion
```

PR #256, opened 2020-03-20
Header style

maryam-mv:


website/src/components/symbols/HeaderImageRight.js
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+const HeaderImageRight = ({ height, width }) => {
+	return (
+		<svg aria-labelledby="title-energy-right" width={width} height={height} viewBox="0 0 268 228">
+			<title>{'Slack'}</title>
```
dominikwilkowski: That's definitely also not a slack svg right? :)
Let's call this like `title-energy-right` and fix the title and `aria-labeldby` id
```suggestion
```

website/src/components/symbols/HeaderImageRight.js
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+const HeaderImageRight = ({ height, width }) => {
+	return (
+		<svg aria-labelledby="title-energy-right" width={width} height={height} viewBox="0 0 268 228">
```
dominikwilkowski:
```suggestion
		<svg width={width} height={height} viewBox="0 0 268 228">
```

website/src/components/symbols/HeaderImageLeft.js
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+const HeaderImageLeft = ({ width, height }) => {
+	return (
+		<svg aria-labelledby="title-energy-left" width={width} height={height} viewBox="0 0 360 228">
```
dominikwilkowski:
```suggestion
		<svg width={width} height={height} viewBox="0 0 360 228">
```

maryam-mv: The aria label and titles were in the existing svg files. Have updated now 👍

website/src/components/symbols/HeaderImageRight.js
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
maryam-mv: It breaks without it so I can only guess that it does need it?

website/src/components/pages/single-component/page-header.js
```diff
@@ -1,30 +1,95 @@
 /** @jsx jsx */
+import React, { useEffect, useState } from 'react';
```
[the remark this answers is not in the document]
maryam-mv: This one breaks without it.

website/src/components/pages/single-component/_utils.js
```diff
@@ -139,3 +139,27 @@ const IconTitle = ({ icon: Icon, children }) => {
...
+			backgroundColor = `linear-gradient(to right, ${colors.hero} 0%, #00468e 50%, #00adbd 100%)`;
+			break;
+		case 'BOM':
+			backgroundColor = colors.hero;
+			break;
+		case 'BTFG':
+			backgroundColor = colors.hero;
+	}
+	return backgroundColor;
+};
```
dominikwilkowski: An easier way to do this would be:
```js
const styleMap = {
  WBC: COLORS.primary,
  WBG: COLORS.primary,
  BSA: `linear-gradient(to right, ${COLORS.hero} 0%, #00468e 50%, #00adbd 100%)`,
  ...
};
```
and then just call it with your brand:
```sh
css={{
  ...styleMap(BRAND)
}}
```

website/src/components/pages/single-component/component-page-header-image.js
```diff
@@ -0,0 +1,91 @@
+/** @jsx jsx */
+import { jsx } from '@westpac/core';
+import React from 'react';
+import HeaderImageRight from '../../symbols/WbcBackgroundRightSvg';
+import HeaderImageLeft from '../../symbols/WbcBackgroundLeftSvg';
+import StgHeaderBackground from '../../symbols/StgBackgroundSvg';
+import BsaHeaderBackgroundSvg from '../../symbols/BsaBackgroundSvg';
```
dominikwilkowski: You did ask about where to put these. I would handle them the same as any other component. Can just be in the `website/src/components` folder... Maybe to make it more obvious we can stick them into `website/src/components/header`?

PR #261, opened 2020-03-25
Admin ui fixes

MadeByMike:
Fixes a few minor UX issues in the Admin UI block editor.


website/field-content/src/views/editor/add-block.js
```diff
@@ -162,6 +162,7 @@ let AddBlock = ({ editorState, editor, blocks }) => {
 							padding: 0,
 							margin: 0,
 							border: 'solid 1px #eaeaea',
+							zIndex: '9999999999999',
```
dominikwilkowski: 🤦‍♂

website/src/components/a11y-filter.js
```diff
@@ -4,7 +4,12 @@ import { jsx, useBrand } from '@westpac/core';
 import Select from 'react-select';
 
 const SVGFilters = () => (
-	<svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" version="1.1">
+	<svg
+		aria-hidden="true"
+		style={{ display: 'none' }}
```
dominikwilkowski: did you cross browser check this? I think adding `display:none` to svgs breaks safari or firefox?

website/field-content/src/views/editor/add-block.js
MadeByMike: Sorry not sorry

website/src/components/a11y-filter.js
MadeByMike: I did not /honesty

MadeByMike: Dom it pains me to say it but you are correct!

MadeByMike: Fix on the way

website/field-content/src/views/editor/add-block.js
MadeByMike: There is a bit of "whack a mole" with z-indexes at the moment. We should revisit this later

PR #263, opened 2020-03-25
Admin ui fixes

maryam-mv:


website/src/components/a11y-filter.js
```diff
@@ -108,7 +108,7 @@ const VisionFilter = ({ children }) => {
 			<SVGFilters />
 
 			<Select
-				css={{ zIndex: 99999 }}
+				css={{ zIndex: 4 }}
```
dominikwilkowski: This is a notable improvement if you ask me 🎉

PR #265, opened 2020-03-25
Refactor FormCheck internal ID allocation

jonnystening:
Refactor internal ID allocation and add instanceIdPrefix prop, as per other form based components

@jaortiz can you please look into the use of context in this component. Should children be passed props, or can Option leverage some context?


components/form-check/src/FormCheck.js
```diff
@@ -106,29 +117,34 @@ export const FormCheck = ({
 			allChildren.push(
 				<Option
 					key={index}
+					index={index}
 					value={props.value}
 					checked={props.checked || checked.includes(props.value)}
 					handleChange={handleChange}
 					type={type}
 					name={name}
+					name={name}
```
jaortiz: Remove duplication

jaortiz:
Context use is fine here. My belief is that you still pass a component the props it needs to work. Passing of the state object in context is for children to have context of what is happening at the root component for overrides stuff.

components/form-check/src/FormCheck.js
jonnystening: Hmm, that was weird. Failed auto merge? 
Fixed

PR #270, opened 2020-03-26
Deploy branch

dominikwilkowski:


website/index.js
```diff
@@ -1,4 +1,5 @@
-require('dotenv').config();
```
dominikwilkowski: What about the `.env` file for the other configs like cloudinary keys?

website/migrations/20200325003307_initial-tables.js
```diff
@@ -0,0 +1,178 @@
+const path = require('path');
```
dominikwilkowski: Should this file be git tracked?

website/migrations/20200325005855_initial-users.js
```diff
@@ -0,0 +1,30 @@
+const path = require('path');
+const migrationName = path.basename(__filename);
+
+const users = [
+	{ email: 'john@thinkmill.com.au', hash: '$2y$10$YU4x2JB04tDN2HED9XKm6OIljAlKoZjaEx7Ip2z/mYiXjR5sT8k.e' },
+	{ email: 'dominik@thinkmill.com.au', hash: '$2y$10$4HRe4RVVJP8NcPxFh/8sWu7z/AaqW4CaYgPA3SN.tPHscQS1uNpCy' },
+	{ email: 'mike.r@thinkmill.com.au', hash: '$2y$10$UVFHfDjIcNEl5pFMKHcsQehwt5yeQrNkvQxHVN5kOqsz1JhKZ0l3G' },
```
dominikwilkowski: Even thought they are hashes I still feel funny this is in the git history

website/index.js
molomby: Not sure what you mean? All the env vars should be in the `config.env`, as they are in dev. (Just the file is in a different spot in production)

website/migrations/20200325003307_initial-tables.js
molomby: Lol , it's [`path`](https://nodejs.org/api/path.html), not `./path` 😄

dominikwilkowski: I mean the file... not this specific line :D

website/migrations/20200325005855_initial-users.js
molomby: You should change your password on first sign in. It still doesn't feel great but I think it's reasonable.

website/migrations/20200325003307_initial-tables.js
molomby: Aha, sorry.

You mean, should we have the migration that creates the tables in git? Sure, why not? It makes it pretty easy to setup a dev environment. Or a prod environment for that matter...

PR #274, opened 2020-03-27
Block editor UI polish

jossmac:
- move some UI around
- align block-editor styles with keystone's native styles (color, spacing, etc.)
- verbose styled-component names to aid understanding
- find and implement icons for the insert menu
- replace "selection-style" toolbar with a more traditional toolbar, that's sticky
- move to absolute offsets for menus, so we can portal (helps with stacking context issues)
- support "Enter" on dynamic-components to create a new paragraph below
- remove stacking context from editor DOM node
- maintain disclosure line on blocks in edit mode, even when not focused
- a little trickery to hide menus when not focused


website/index.js
```diff
@@ -42,7 +42,7 @@ const apps = [
 	new GraphQLApp(),
 	new AdminUIApp({
 		adminPath: '/admin',
-		authStrategy,
+		// authStrategy,
```
jossmac: TODO: Uncomment after WIP

website/utils/dynamic-components-view.js
```diff
@@ -36,27 +41,26 @@ export let Provider = ({ value, children }) => {
 export function Sidebar({ editor }) {
 	let { adminMeta, components } = useContext(Context);
 	let [view] = adminMeta.readViews([components]);
-	const icon = (
```
jossmac: This was **always** rendering an image icon.

website/index.js
MadeByMike:
```suggestion
		authStrategy,
```

MadeByMike:
@jossmac I removed the comment around the authStrat - are you happy to move this out of WIP?

jossmac:
No. Not happy to move out of WIP. There's lots of things that aren't finished.

PR #275, opened 2020-03-29
A11y retesting fixes

jonnystening:


components/button/package.json
```diff
@@ -32,7 +32,8 @@
 		"@westpac/core": "^1.0.0",
 		"@westpac/hooks": "^1.0.0",
 		"@westpac/icon": "^1.0.0",
-		"prop-types": "^15.7.2"
+		"prop-types": "^15.7.2",
+		"react-router-dom": "^5.1.2"
```
dominikwilkowski: That can't be right!

dominikwilkowski: This has to go into the devDependencies or we ship this with the component

jaortiz: Odd, I ran the flag to add it to devDependencies but it seems like it didn't work...

components/breadcrumb/src/Crumb.js
```diff
@@ -67,7 +67,9 @@ Crumb.propTypes = {
 	current: PropTypes.bool,
 
 	/**
-	 * An href for a link
+	 * The crumb link href value.
+	 *
+	 * Defaults to '#0'.
```
dominikwilkowski, line 72: What do you mean "defaults to '#0'? There is no defaultProp declared tho

components/breadcrumb/src/overrides/link.js
```diff
@@ -33,7 +26,10 @@ const linkStyles = (_, { current }) => {
 	};
 };
 
-const linkAttributes = (_, { current, href }) => ({ href: current ? null : href });
+const linkAttributes = (_, { current, href }) => ({
+	href: href || '#0',
```
dominikwilkowski, line 30: Can we make this a default prop?
Also what is `#0`?

components/form-check/examples/30-disabled-state.js
```diff
@@ -37,7 +37,7 @@ function Example({ brand }) {
 				</Option>
 			</FormCheck>
 
-			<FormCheck type="radio" name="example-radio-medium-disabled" defaultValue="2">
```
dominikwilkowski, line 40: We added those default values to see what the select state looks like from a design perspective. If that's not a11y we should not advertise it but not remove it from the tests

components/input-group/examples/40-overrides.js
```diff
@@ -93,31 +98,32 @@ function Example({ brand }) {
...
+
 			<h2>With overrides and component overrides</h2>
 			<InputGroup
+				name="example-overrides-component-1"
 				overrides={{
 					Button: {
 						styles: styles => ({
 							...styles,
-							outline: '3px dotted blue',
+							outline: '4px solid blue',
```
dominikwilkowski, line 110: Why are we making changes to the overrides tests? I don't understand in the context of a11y changes? What's the context here I'm missing?

components/input-group/src/Right.js
```diff
@@ -4,18 +4,18 @@ import { jsx } from '@westpac/core';
 import PropTypes from 'prop-types';
 
 import { Button } from './Button';
-import { Select } from './Select';
-import { Label } from './Label';
+import { SelectField } from './SelectField';
+import { Text } from './Text';
```
dominikwilkowski, line 8: ?

components/breadcrumb/src/Crumb.js
jonnystening: Defaulted in code

components/breadcrumb/src/overrides/link.js
jonnystening: `href="#0"` is a nice default href value to use, like the common `href="#"` but with the `0` it stops the scroll jump out of the box without requiring any js wizardry. 

`<a>` must have a `href` attribute in order to get assistive technology support (without having to muck around with tabindex etc).

components/input-group/examples/40-overrides.js
jonnystening: I just wanted to make the style overrides more obvious

components/input-group/src/Right.js
jonnystening: `SelectField` renders our `<select>` and `<label>` add-on

`Text` is our text add-on

Question?

components/form-check/examples/30-disabled-state.js
jaortiz: @dominikwilkowski I had to remove the defaultValue in these instances because it was set to an option that was disabled which made the entire radio-group untabbable to making it impossible to reach to a keyboard user.

PR #277, opened 2020-03-30
Replace `<>` shorthands with `<React.Fragment>`

molomby:
Work around for `transform-react-jsx: pragma has been set but pragmaFrag has not been set` error (see https://github.com/zeit/next.js/issues/11230)


website/src/components/playground/macro.js
```diff
@@ -77,7 +77,7 @@ function transformPlayground({ path, identifierName, babel, state }) {
 
 			if (path.node.children.length > 1) {
 				//Add children to a fragment
-				code = `<>${code}</>`;
+				code = `<React.Fragment>${code}</React.Fragment>`;
```
dominikwilkowski: Can we please make this a `<Fragment>` to stay just a little sane :)

PR #279, opened 2020-03-30
component page anchor, page-content dynamic-block style, and sticky header fix.

maryam-mv:


website/src/components/header/component-page-header-image.js
```diff
@@ -2,7 +2,7 @@
 
 import { Fragment } from 'react';
 import { jsx } from '@westpac/core';
-
+import React, { Fragment } from 'react';
```
dominikwilkowski: You got from react important twice here

PR #296, opened 2020-04-02
Improve ui when navigating to different pages

maryam-mv:
- removed data (useQuery(ALL_PAGES)) that was being fetched from layout and passed down to navigation but never used.
- moved up navigation data (items in the navigation) logic to the layout component
- moved the footer component to the dynamic [...page]
- removed all unnecessary loading ui
- add dynamic routes to Link

Now the page doesn't flash to blank displaying a "Loading ..." UI 🎉

Note, this could still use some improvement as the data seems to still refetch when it doesn't need to.


website/src/components/layout/navigation/navigation.js
```diff
@@ -2,31 +2,17 @@
...
-import gql from 'graphql-tag';
 
 import { jsx, useBrand } from '@westpac/core';
 import { Heading } from '@westpac/heading';
-import { useQuery } from '@apollo/react-hooks';
 import { NavigationBlock } from './navigation-block';
 
-export const Navigation = ({ components }) => {
-	const { SPACING } = useBrand();
+// root paths that don't need to be dynamic
```
maryam-mv, line 10: Added this PATHS array for root pages like home and tokens. This is assuming that all other items will have children.

maryam-mv: The alternative would have been to add a field in the settings with the correct path. This was avoided as the settings JSON file is already not the easiest to edit and work with in the admin ui. We also wanted to avoid making routing even more of a responsibility for authors.

PR #300, opened 2020-04-03
fix icon label overflow bug

maryam-mv:
Quick fix, span text is more condensed than <code />

### DIFF
--- a/website/dynamic-blocks/icons.js
+++ b/website/dynamic-blocks/icons.js
@@ -38,7 +38,7 @@ const renderIcons = search => {
 							padding: 5,
 						}}
 					/>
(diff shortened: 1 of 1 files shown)


maryam-mv:
> Wouldn't it be better to just add `overflow: auto`?
> I like the semantic `<code>` tag usage here

overflow auto will just hide the overflowing text. I like code as well but the letter spacing/ width makes it hard to fit in without making it really small. What are your thoughts? I thought the regular text looked fine :) Up to you

maryam-mv:
@dominikwilkowski  looks like this: 
Screen Shot 2020-04-06 at 11 09 29 am

dominikwilkowski:
Makes sense!

PR #304, opened 2020-04-06
Tabcordion callbacks

jaortiz:
@dominikwilkowski I've only added onOpen and onClose callback props here. We only animate in accordion mode so wasn't sure if the opening and closing callbacks were useful. Thoughts?


components/tabcordion/examples/50-callbacks.js
```diff
@@ -0,0 +1,63 @@
+/** @jsx jsx */
+
+import { GEL, jsx } from '@westpac/core';
+import { Tab, Tabcordion } from '@westpac/tabcordion';
+
+function Example({ brand }) {
+	const onOpen = () => {
+		console.log('This function is run when a tab is opened');
```
dominikwilkowski, line 8: How do you know which tab was opened?

dominikwilkowski:
For a more applied example:

Imagine we use the tabcordion to show and hide parts of the page like like the `Design` or `Code` tabs here: https://gel.test.do.westpac.thinkmill.cloud/components/alert
If we wanted to make sure the url is altered for each tab so someone could link to a tab we would require a callback that is also aware of the tab that is open.

Hope this helps to frame this PR

PR #307, opened 2020-04-07
Fix/ui bugs

maryam-mv:
- fix admin ui not clickable images
- fix header spacing
- make header responsive
- update stg header colour
- remove version if version value is undefined
- page-content component UI style fixes


website/dynamic-blocks/image.js
```diff
@@ -84,8 +84,9 @@ export const Image = {
 			</Fragment>
 		);
 	},
-	component: ({ caption, image }) => {
+	component: ({ caption, image, context }) => {
 		const { SPACING } = useBrand();
+		console.log(context);
```
dominikwilkowski, line 89:
```suggestion
```

PR #310, opened 2020-04-08
Add shortcodes

MadeByMike:
Add shortcodes:

image

image

Configure here:

image


website/src/shortcodes.js
```diff
@@ -0,0 +1,5 @@
+import { brandsMap } from './components/brand-switcher/brand-switcher';
+
```
MadeByMike, line 2:
```suggestion
// Shortcodes can contain characters a-z and numbers 0-9 but no other characters
```

PR #312, opened 2020-04-09
Migration to new Keystone relations

MadeByMike:
Note: This cannot be merge until db backups exist and more testing is done.

In this branch: 

- Updates Keystone version
- Change the way related content is done
- Fix to related content links
- Change to the way navigation URLs are generated
- Remove the virtual field `name` and remove reliance on this everywhere in the front-end

Related Pages will need to be re-created.


dominikwilkowski:
Fixed your base 😬

MadeByMike:
> I don't love that there's no rollback but I guess it's not even really live yet, right?

Correct! Rollback will be to restore the database dump.

@molomby could you please endorse me for SQL on linkedIn?

MadeByMike:
Ok I have tested the migration with the staging database and this went off without a hitch! \o/

Unfortunately the rest of syncing this branch with develop did not. There were merge conflict resolutions gone wrong which had resulted in some loss of some URL updates. 

And components updated **cough** vision filters **cough** @jaortiz, that expected new data for headings etc... when this new data was not found on existing pages it caused them to crash. I solved this by adding default values but we should checkin with everyone on the process to avoid this in future.

The final thing is we didn't totally get the navigation and related page links right every-time when using Next.js dynamic links so I reconfigured around this a little.

Now... we should be able to merge this and run the knex migration in staging.

PR #314, opened 2020-04-09
WIP: Fix page content

jaortiz:


website/dynamic-blocks/intro-section.js
```diff
@@ -54,11 +54,13 @@ const TableOfContents = () => {
 		const headings = [];
 		if (containerRef && containerRef.current) {
 			const container = containerRef.current.closest('.slate-container') || document;
-			container.querySelectorAll('[data-toc]').forEach((h, i) => {
-				headings[i] = <TableLink key={`nav-${i}`} headingId={h.id} headingText={h.innerText} />;
-			});
+			setTimeout(() => {
```
jaortiz, line 57: @dominikwilkowski The page-content list wasn't being shown because the headings are rendered after this has run so it can't find them. Not sure of the best way to solve this.

website/dynamic-blocks/intro-section.js
```diff
@@ -54,11 +54,13 @@ const TableOfContents = () => {
...
+				container.querySelectorAll('[data-toc]').forEach((h, i) => {
+					headings.push(<TableLink key={`nav-${i}`} headingId={h.id} headingText={h.innerText} />);
+				});
+				setToc(headings);
+			}, 100);
```
jaortiz, line 62: @dominikwilkowski The page-content list wasn't being shown because the headings are rendered after this has run so it can't find them. Not sure of the best way to solve this.

dominikwilkowski: @MadeByMike Maybe you got an idea?

MadeByMike: The headings were changed and the `data-toc=true` is no longer there. Therefor the querySelector isn't working. 

There may or may not be an issue with render order, but from what I can tell I don't think this change will fix it. Does it work @jaortiz ?

jaortiz: @MadeByMike The `data-toc` are added to headings generated by the headings block I recently created if the option is selected to include it in the table of contents and I can see that it is working correctly. 

The timeout fix works and the page content links are being rendered properly so it seems to be a render order issue. Just wanted to know if there was a better way of solving this though?

PR #321, opened 2020-04-15
Update toc

MadeByMike:
- New method for generating the toc that is SSR compatible.
- Removes old heading component from toolbar

BONUS: Also fixes some React warnings in console.

Closes #314


jaortiz:
Awesome fix!

PR #323, opened 2020-04-16
Url addressable tabs

maryam-mv:
Note that there are some changes to the Tabcordion component here. I've updated the onOpen function to use index instead of tabId to make it more usable by client.


components/tabcordion/src/Tab.js
```diff
@@ -135,7 +137,7 @@ export const Tab = forwardRef(
 		useEffect(() => {
 			if (mode === 'tabs') {
 				if (selected) {
-					onOpen(tabId);
+					onOpen(idx);
 				} else {
 					onClose(tabId);
 				}
```
jaortiz, line 143: We should probably pass both idx and tabId to all of the callback functions for consistency

dominikwilkowski: yeah let's do both

dominikwilkowski: I'd do that as an object and favor `idx` over `tabId` as it's more specific.

components/tabcordion/src/Tab.js
```diff
@@ -135,7 +137,7 @@ export const Tab = forwardRef(
 		useEffect(() => {
 			if (mode === 'tabs') {
 				if (selected) {
-					onOpen(tabId);
+					onOpen({ idx, tabId });
 				} else {
 					onClose(tabId);
```
jaortiz, line 142: We should pass the idx and tabId to onClose, onOpening and onClosing to be consistent

jaortiz:
LGTM

PR #329, opened 2020-04-17
wrap with conditional

maryam-mv:
Sorry quick fix.

### DIFF
--- a/website/dynamic-blocks/headings.js
+++ b/website/dynamic-blocks/headings.js
@@ -34,12 +34,13 @@ const Component = ({ heading, size, level, addTableContent, subText, text }) => 
 					</Cell>
 				</Grid>
 			</WestpacHeading>
-
-			<Grid columns={12}>
-				<Cell width={10} left={[1, 1, 2, 2]}>
-					{subText && text && <p css={{ lineHeight: 2 }}>{text}</p>}
-				</Cell>
-			</Grid>
+			{subText && text && (
+				<Grid columns={12}>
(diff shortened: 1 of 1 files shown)


website/dynamic-blocks/headings.js
```diff
@@ -34,12 +34,13 @@ const Component = ({ heading, size, level, addTableContent, subText, text }) =>
 					</Cell>
 				</Grid>
 			</WestpacHeading>
-
-			<Grid columns={12}>
-				<Cell width={10} left={[1, 1, 2, 2]}>
-					{subText && text && <p css={{ lineHeight: 2 }}>{text}</p>}
-				</Cell>
-			</Grid>
+			{subText && text && (
```
dominikwilkowski, line 37: what's `subText` and why is it required for the blow below?

jaortiz: Kate asked for the subText to be conditionally shown.

dominikwilkowski: We we don't display `subText` anywhere below? is that a checkbox somewhere?

maryam-mv: Yeah subText is a boolean that comes from a checkbox.

dominikwilkowski: Makes sense then :) thank you both

maryam-mv: Screen Shot 2020-04-17 at 1 14 13 pm

PR #332, opened 2020-04-19
Add left indent back

MadeByMike:
Add left indent back

Causing error will investigate cause later


maryam-mv:
I don't think we want to add the indent back. It needs to use the Grid and be in line with other text on the page as per design. I'm happy to to take care of this :)

PR #334, opened 2020-04-20
Sticky tabs and footer

jaortiz:


jaortiz:
@dominikwilkowski Requirement for the footer were to make it act the same as current GEL site where it gets fixed to the bottom when the user scrolls halfway down the page and disappears when they scroll up past halfway hence all the logic. So its not really a sticky footer :stuck_out_tongue:

dominikwilkowski:
I see. All good then :)

PR #335, opened 2020-04-20
Nav ui updates

maryam-mv:


website/src/components/brand-switcher/brand-switcher.js
```diff
@@ -71,8 +71,8 @@ export const BrandSwitcher = () => {
 
 	return (
 		<Fragment>
-			<div css={{ height: 35, padding: SPACING(2) }}>
-				<Link href={`/?b=${brandName}`}>
+			<div css={{ height: 35, padding: `${SPACING(6)} ${SPACING(4)}` }}>
+				<Link href={`/design-system?b=${brandName}`}>
```
MadeByMike, line 75: These links won't work. Next needs to link to a page template. Should be `<Link href={BASE_PAGE} as={`${BASE_URL}?b=${brandName}`}>`

dominikwilkowski, line 75:
```suggestion
				<Link href={BASE_PAGE} as={`${BASE_URL}?b=${brandName}`}>
```

MadeByMike: @dominikwilkowski you will need to add these imports as well, not sure those variables are in scope

PR #363, opened 2020-04-28
component page shell updates

jaortiz:


website/src/components/symbols/StgBackgroundSvg.js
```diff
@@ -1,13 +1,15 @@
...
+
 import { jsx } from '@westpac/core';
+import { Svg } from './Svg';
 
 export const StgBackgroundSvg = ({ width, height }) => {
 	return (
-		<svg
-			aria-labelledby="stgeorge-header-image"
+		<Svg
+			aria-label="stgeorge-header-image"
```
dominikwilkowski, line 9: Is that a good label that is read out? Should that be human readable?

PR #365, opened 2020-04-29
fix jason parse bug

maryam-mv:
Sorry about that 😬

### DIFF
--- a/website/src/components/_utils.js
+++ b/website/src/components/_utils.js
@@ -65,7 +65,7 @@ export const RelatedInformation = ({ item }) => {
 	try {
 		hasRelatedInfo = checkNode(JSON.parse(relatedInfo.document));
 	} catch (e) {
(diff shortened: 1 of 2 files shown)


website/src/components/_utils.js
```diff
@@ -61,7 +61,12 @@ export const RelatedInformation = ({ item }) => {
 	function checkNode(node) {
 		return node.text || (node.nodes && node.nodes.some(checkNode));
 	}
-	const hasRelatedInfo = checkNode(JSON.parse(relatedInfo.document));
+	let hasRelatedInfo = false;
+	try {
+		hasRelatedInfo = checkNode(JSON.parse(relatedInfo.document));
+	} catch (e) {
+		console.info('Could not parse document data searching for Related Info');
```
dominikwilkowski, line 68: should likely be a warn
```suggestion
		console.warn('Could not parse document data searching for Related Info');
```

PR #376, opened 2020-04-30
Accessibility retest #2

jonnystening:


components/breadcrumb/examples/00-default.js
```diff
@@ -10,16 +10,25 @@ function Example({ brand }) {
...
+			<h3>As links</h3>
 			<Breadcrumb>
 				<Crumb href="#/" text="Home" />
 				<Crumb href="#/personal-banking/" text="Personal" />
 				<Crumb href="#/credit-cards/" text="Credit cards" />
 			</Breadcrumb>
 
-			<hr />
-
-			<h2>Declarative as router</h2>
```
dominikwilkowski, line 22: It would be good to see the `tag` prop in the examples somewhere

jonnystening: yes, could I get either you @dominikwilkowski or you @jaortiz to assist?

components/breadcrumb/examples/00-default.js
```diff
@@ -5,6 +5,9 @@ import { Breadcrumb, Crumb } from '@westpac/breadcrumb';
 
 import { Intopia } from '../../../helpers/example/components/Intopia.js';
 
+// Crumb tag as a component example
+const CrumbTag = props => <a {...props} onClick={() => console.log(`Clicked crumb link`)} />;
```
jonnystening, line 9: Added an example showing `tag` prop taking a component...

components/breadcrumb/src/Crumb.js
```diff
@@ -62,22 +69,25 @@ export const Crumb = ({ current, href, text, onClick, overrides, ...rest }) => {
...
 
 	/**
 	 * The text of the crumb
 	 */
 	text: PropTypes.string.isRequired,
 
+	/**
+	 * Component tag, may receive a react-router Link
+	 */
+	tag: PropTypes.node,
```
jonnystening, line 89: @dominikwilkowski `PropTypes.node` give me a `checkPropTypes.js:20 Warning: Failed prop type: Invalid prop `tag` supplied to `Crumb`, expected a ReactNode.` 🤔

dominikwilkowski: Yeah cause your defaultProp is a string.... You should make that a component:
```
const Anchor = (props) => <a {...props} />
```

dominikwilkowski: To be clear: you shouldn't use defaultProps here but check if the props is set and then override the anchor. Using a defaultProp here as a string has to fail as we don't want people to just add strings.

jonnystening: First on my list was to ask you how to set a default setting a component as you had mentioned

Will set up in code, please hold, will commit shortly

jonnystening: Cos I can’t make the value a component (function) and PropTypes.node can be anything... incl string

PR #379, opened 2020-05-04
Adding cookie secret config to Keystone init

molomby:
This needs to be set to make the app production ready.

I've added config to the live and staging servers setting this value so all we need to do is roll this change out. All sessions will be reset (once) when this change is deployed (but after that will be stable).

Developers will need to add a `COOKIE_SECRET` value to their `.env` files, as per the notes in `website/example.env`.


molomby:
Related https://github.com/keystonejs/keystone/issues/2880

PR #381, opened 2020-05-05
homepage header ui updates

maryam-mv:


website/src/components/header/home-page-header-image.js
```diff
@@ -18,23 +18,23 @@ const HomePageHeaderImage = ({ brand }) => {
 				},
 			}}
 		>
-			<WbcBackgroundRightSvg height={'100%'} />
+			<WbcBackgroundRightSvg height={'629px'} />
```
dominikwilkowski, line 21: We need to use the SPACING tokens here.
629px = 39.3125rem = 105units which is close with 39.375rem.

So should be:
```suggestion
			<WbcBackgroundRightSvg height={SPACING(105)} />
```

maryam-mv: Hey Dom, are you sure SPACING is the semantically correct token to use here for a size property? I also kinda need this to be exactly 629px and SPACING(105) would give me 630 px (6 * 105).

dominikwilkowski: The `SPACING` token is exactly meant for vertical spacing between things.
The only place we should not use it is in padding or margin, border-width etc really.

jaortiz: So we shouldnt be using SPACING in paddings or margins? Cause Ive been using it for that when I can.

dominikwilkowski: not when you end up with `SPACING(0.0125)` or something... 😬

PR #382, opened 2020-05-06
added blender support

dominikwilkowski:


components/button-dropdown/src/overrides/heading.js
```diff
@@ -1,25 +1,27 @@
...
+	const { COLORS, TYPE } = useBrand();
 
 	return {
-		margin: '1.5rem -0.375rem 1.125rem -0.375rem',
-		padding: '0.4375rem',
-		backgroundColor: COLORS.background,
-		color: COLORS.hero,
-		fontWeight: '500', // what is the correct way to do this??
-		// fontSize: '0.75rem',
+		'&&&': {
```
jaortiz, line 14: Why are we adding this? Are there specificity issues?

dominikwilkowski: yes specificity needs to be overridden from the Heading component

PR #390, opened 2020-05-11
fixed grid up with examples and added test layout

dominikwilkowski:


dominikwilkowski:
To close this off; the above work is visible in the grid example `80-test-layout`:
image

I've also created the same layout in the old grid here:
image

Download here:
[GUI-blend.zip](https://github.com/WestpacGEL/GEL/files/4607449/GUI-blend.zip)

PR #391, opened 2020-05-11
WIP: Refactor breakpoints

jonnystening:


jonnystening:
Good news @dominikwilkowski we're keeping the new 5th breakpoint... just renaming it. Causes the least issues with the platform teams and their legacy codebase. Also helps the designers with language alignment. 

I think it's the best we're going to get on this one sorry 🤷‍♂️

dominikwilkowski:
How are you going to rename an array item?
Like:
```jsx
<Component css={{
  padding: [2,20,40,80],
}} />
```

Just curious...

dominikwilkowski:
I was thinking that it would be better and more explicit (but more typing) if we made it an object.
So going from:
```jsx
<Component css={{
  padding: [2,20,40,80],
}} />
```
Into this:
```jsx
<Component css={{
  padding: {
    xs: 2,
    sm: 20,
    md: 40,
    lg: 80,
  },
}} />
```

That way you can also skip things easier:
```jsx
<Component css={{
  padding: {
    xs: 2,
    lg: 80,
  },
}} />
```
or even include breakpoints that are not in the layout tokens:
```jsx
<Component css={{
  padding: {
    xs: 2,
    md: 40,
    "max-width: 1400px": 80,
  },
}} />
```

jonnystening:
> How are you going to rename an array item?
> Like:
> 
> ```js-jsx
> <Component css={{
>   padding: [2,20,40,80],
> }} />
> ```
> 
> Just curious...

We don't. They stay the same :). It's just the name we use to refer to it that changes.

jonnystening:
> I was thinking that it would be better and more explicit (but more typing) if we made it an object.
> So going from:
> 
> ```js-jsx
> <Component css={{
>   padding: [2,20,40,80],
> }} />
> ```
> 
> Into this:
> 
> ```js-jsx
> <Component css={{
>   padding: {
>     xs: 2,
>     sm: 20,
>     md: 40,
>     lg: 80,
>   },
> }} />
> ```
> 
> That way you can also skip things easier:
> 
> ```js-jsx
> <Component css={{
>   padding: {
>     xs: 2,
>     lg: 80,
>   },
> }} />
> ```
> 
> or even include breakpoints that are not in the layout tokens:
> 
> ```js-jsx
> <Component css={{
>   padding: {
>     xs: 2,
>     md: 40,
>     "max-width: 1400px": 80,
>   },
> }} />
> ```

Yeah that's nice, but a lot of change at this stage. We're just using the facepaint array values feature in emotion which does the trick. The pattern continues to a few props values too, like button size and block props. Can give them array values if needed.

PR #392, opened 2020-05-12
Refactor headings

jonnystening:
**IMPORTANT: Requires a `yarn build` before `yarn start`**

**Packs**
- Renamed the packs `headline` token as `typeScale` and split into two types of packs – for bodyFont and brandFont
- These packs become a way to access the type sizing scale, rather than pre-bundled heading settings (ie. you may use the GEL type scale without it necessarily being a heading)
- fontWeight setting removed from these packs, provided elsewhere for Heading only (headings are not always bold across the brands)
- A larger `1` (60px) option added – bodyFont provides 10 sizes, brandFont provides 7 sizes (input from Justin)
- Line height adjusted to suit the font size (input from Justin)
- Installed fonts for examples

**Type**
- Added a TYPE `headingWeight` token allowing you to set the brand specific default weight for headings (note: could do this inside the `<Heading />` component via a mapping const, but feel this token approach is more visible and easier to configure)
- I put this new token inside TYPE rather than PACKS since it relates directly to the font settings rather than packs of pre-bundled type settings

**Transformer**
- Updated web transformer to output new token values

**Heading**
- Component updated to make use of typeScale and headingWeight tokens
- ~New `fontType` prop added, made it take a string (boolean type wont scale well if/when we add another font set)~
- `<Heading>` and `<BrandHeading>` components exported
- Installed fonts for examples

**Examples**
- Updated to show the thing

**Todo**
- Update `<Heading size={x} />` values now that size 1 is larger


components/core/examples/40-tokens-packs.js
```diff
@@ -9,14 +10,43 @@ function Example({ brand }) {
...
+						<span
+							css={{
+								display: 'block',
+								margin: '0 0 1rem',
+								...PACKS.typeScale.brandFont[i + 1],
+							}}
+						>
+							{i + 1}
+						</span>
+						<Code key={i}>{JSON.stringify(PACKS.typeScale.brandFont[i + 1], null, 2)}</Code>
```
dominikwilkowski, line 30: The key needs to move to the wrapping element, in this case the `Fragment`

components/core/examples/40-tokens-packs.js
```diff
@@ -9,14 +10,43 @@ function Example({ brand }) {
...
+						<span
+							css={{
+								display: 'block',
+								margin: '0 0 1rem',
+								...PACKS.typeScale.bodyFont[i + 1],
+							}}
+						>
+							{i + 1}
+						</span>
+						<Code key={i}>{JSON.stringify(PACKS.typeScale.bodyFont[i + 1], null, 2)}</Code>
```
dominikwilkowski, line 48: Same here. Wrong location for the `key`

components/heading/src/Heading.js
```diff
@@ -60,7 +61,12 @@ Heading.propTypes = {
 	/**
 	 * The visual size of the heading
 	 */
-	size: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9]).isRequired,
+	size: PropTypes.oneOf([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]).isRequired,
```
dominikwilkowski, line 64: This is false because when `fontType` is `"brandFont"` it does not take the last couple of numbers

website/src/components/header/home-page-header.js
```diff
@@ -187,14 +187,12 @@ const HeroIntro = () => {
...
 						size={1}
+						fontType="brandFont"
 						css={mq({
 							paddingTop: [SPACING(4), SPACING(7)],
 							paddingBottom: [SPACING(4), SPACING(6)],
-							fontFamily: `${TYPE.brandFont.fontFamily} !important`,
 							maxWidth: 600,
-							fontWeight: 500,
 							margin: '0 auto !important',
-							zIndex: 3,
```
dominikwilkowski, line 197: Why would the z-index change?

jonnystening: Just not sure why it's there... any idea?

dominikwilkowski: I'd assume cause of the header?

PR #394, opened 2020-05-12
More ui updates

jaortiz:


website/dynamic-blocks/headings.js
```diff
@@ -147,7 +147,7 @@ export const Heading = {
 								Heading: {
 									styles: (styles) => ({
 										...styles,
-										scrollMarginTop: '75px',
+										scrollMarginTop: '166px',
```
jonnystening, line 150: shouldnt this be in rem?

website/dynamic-blocks/vision-filters.js
```diff
@@ -238,7 +238,7 @@ export const VisionFilters = {
 								Heading: {
 									styles: (styles) => ({
 										...styles,
-										scrollMarginTop: '75px',
+										scrollMarginTop: '166px',
```
jonnystening, line 241: rem?

website/src/components/layout/sidebar.js
```diff
@@ -32,9 +32,9 @@ export const Sidebar = ({ items }) => {
 				gridColumnStart: 1,
 				gridColumnEnd: 2,
 				borderRight: `1px solid ${COLORS.border}`,
+				width: '300px',
```
jonnystening, line 35: if we must have this, can we make it '100%' instead?

website/src/components/layout/sidebar.js
```diff
@@ -32,9 +32,9 @@ export const Sidebar = ({ items }) => {
...
 				height: '100vh',
 				boxSizing: 'border-box',
-
```
jonnystening, line 37: 👍

website/src/components/layout/sidebar.js
jaortiz: This is to fix that overhang that I'm getting. I can't figure out how else to hide it. 100% makes it even wider. This is the only way I have found that fixes it.

PR #403, opened 2020-05-18
Refactor breakpoint names

jonnystening:
- Renaming the `sm` (landscape phone and phablet) breakpoint as `xsl` —  which works as either "Extra small landscape" or "Extra small (large)"

- Updated breakpoint name usage and array values notation for components and website


components/alert/examples/00-default.js
```diff
@@ -34,15 +34,17 @@ function Example({ brand }) {
...
-				This alert needs your attention, but it’s not super important. Lorem ipsum dolor sit amet,
-				consectetur adipisicing elit. Quo dolor provident quasi nisi officia tempore fuga autem,
-				animi iste molestiae, qui omnis doloribus aliquid ipsam rem fugiat veniam voluptatem
-				accusamus! Lorem ipsum dolor sit amet, consectetur adipisicing elit. Est, unde quis,
-				molestias nisi quae voluptates nemo quaerat nihil, consequuntur nobis ratione rerum
-				asperiores eveniet dicta maiores quia nostrum. Pariatur, natus. Lorem ipsum dolor sit amet,
-				consectetur adipisicing elit. Fuga, magnam illum harum consequatur, quo sunt impedit quam
-				minus eaque saepe voluptas corrupti voluptatum, sapiente dolor sequi tempore maxime? Neque,
-				obcaecati. <a href="#">Link</a>
+				<p>
```
dominikwilkowski, line 37: You shouldn't have to do this in the alert components

helpers/example/components/AppStart.js
```diff
@@ -42,7 +42,6 @@ export function AppStart({ components, packageName, pkg, version }) {
 							`}
 						/>
 						<Sidebar components={components} brand={brand} setBrand={setBrand} />
-						<p>{version}</p>
```
dominikwilkowski, line 45: Why remove the version?

jonnystening: It’s in there twice (already in the page heading), that one is breaking the page layout

components/alert/examples/00-default.js
jonnystening: May be used with multiple paragraphs 🤷‍♂️

helpers/example/components/AppStart.js
dominikwilkowski: uhm not sure what you mean but the version is not duplicated anywhere I can find???

jonnystening: Screen Shot 2020-05-19 at 10 37 25 am

dominikwilkowski: got you

PR #404, opened 2020-05-19
Fix modal

jonnystening:
- wrap Modal with a GEL wrapper: Modal uses portal so doesn't get a GEL wrapper, and therefore misses out on some styling (e.g. close button focus styling)

- tidy up


components/modal/src/Modal.js
```diff
@@ -114,36 +115,38 @@ export const Modal = ({
 	if (typeof window !== 'undefined') {
 		return ReactDOM.createPortal(
 			<ModalContext.Provider value={{ state }}>
```
dominikwilkowski, line 117: I would wrap everything with `<GEL/>`. No reason to leave the ModalContext provider outside that I can see.
Having `<GEL/>` outside will be more consistent with everything else we do but in the end it won't make much difference.

PR #406, opened 2020-05-19
Refactor grid

jonnystening:
- Updates Grid's `rowGap` propType declaration (should be able to take string value - rowGap's default CSS value is 'normal')

- Updates Cell's `left` propType declaration (should be able to take string value – grid-column-start's default value is 'auto')

Also includes a fix for GEL `brand` prop's incorrect propType declaration


jonnystening:
> Looks fine but wouldn't you just leave it empty if you wanted the default?

Thanks for the late review mate appreciate it's late.

If you use responsive (array) values you need to be able to switch back to the default `<Cell left={[2, 'auto' ]}>`

PR #415, opened 2020-05-26
Responsive Heading size

jonnystening:
Gives the `size` prop the ability to take responsive array value

If `tag` is not provided it will take the first array index value


components/heading/src/overrides/brandHeading.js
```diff
@@ -25,12 +27,29 @@ const BrandHeading = forwardRef(({ state: { tag: Tag, size }, ...rest }, ref) =>
 
 const brandHeadingStyles = (_, { size }) => {
 	const { PACKS, TYPE } = useBrand();
+	const mq = useMediaQuery();
+
+	const tokens = PACKS.typeScale.brandFont;
+	const sizeMax = Object.keys(tokens).reduce((a, b) => (tokens[a] > tokens[b] ? a : b));
```
jaortiz, line 33: This is doing a comparison on the token size object which doesn't make sense. Should just be 
`Math.max(...Object.keys(tokens))` to find the max key value

components/heading/src/overrides/brandHeading.js
```diff
@@ -25,12 +27,29 @@ const BrandHeading = forwardRef(({ state: { tag: Tag, size }, ...rest }, ref) =>
...
+	let sizeArr = asArray(size);
+
+	// check for invalid sizes
+	sizeArr = sizeArr.map(function (s) {
```
jaortiz, line 37: Use an arrow function here for consistency

components/heading/src/overrides/heading.js
```diff
@@ -25,12 +27,29 @@ const Heading = forwardRef(({ state: { tag: Tag, size }, ...rest }, ref) => {
 
 const headingStyles = (_, { size }) => {
 	const { PACKS, TYPE } = useBrand();
+	const mq = useMediaQuery();
+
+	const tokens = PACKS.typeScale.bodyFont;
+	const sizeMax = Object.keys(tokens).reduce((a, b) => (tokens[a] > tokens[b] ? a : b));
```
jaortiz, line 33: Same comment as above

components/heading/src/overrides/heading.js
```diff
@@ -25,12 +27,29 @@ const Heading = forwardRef(({ state: { tag: Tag, size }, ...rest }, ref) => {
...
+	let sizeArr = asArray(size);
+
+	// check for invalid sizes
+	sizeArr = sizeArr.map(function (s) {
```
jaortiz, line 37: Arrow function here please

PR #443, opened 2020-06-11
WIP: Feature/blender support

jaortiz:


components/switch/blender/recipe.js
```diff
@@ -0,0 +1,166 @@
...
+	return (
+		<GEL brand={brand}>
+			<Switch name="example-small" size="small" label="example" />
+			<Switch name="example-small" size="small" label="example" />
+			<Switch name="example-small-checked" size="small" label="example" checked />
+			<Switch name="example-small-block" size="small" label="example" block />
+			<Switch name="example-small-block-disabled" size="small" label="example" block disabled />
+			<Switch name="example-small-block-checked" size="small" label="example" block checked />
+			<Switch name="example-small-disabled" size="small" label="example" disabled />
+			<Switch name="example-small-disabled-checked" size="small" label="example" disabled checked />
```
dominikwilkowski, line 16: You can greatly simplify this by doing `name="TEXT"` everywhere instead

components/switch/blender/recipe.js
```diff
@@ -0,0 +1,166 @@
+import { GEL } from '@westpac/core';
+import React from 'react';
+
+import { Switch } from '@westpac/switch';
+
+export function AllStyles({ brand }) {
+	return (
+		<GEL brand={brand}>
+			<Switch name="example-small" size="small" label="example" />
+			<Switch name="example-small" size="small" label="example" />
```
dominikwilkowski, line 10: why is this twice in there?

components/switch/blender/recipe.js
```diff
@@ -0,0 +1,166 @@
...
+			<Switch name="example-small-checked" size="small" label="example" checked />
+			<Switch name="example-small-block" size="small" label="example" block />
+			<Switch name="example-small-block-disabled" size="small" label="example" block disabled />
+			<Switch name="example-small-block-checked" size="small" label="example" block checked />
```
dominikwilkowski, line 14: What about `block checked disabled` combination?

components/switch/blender/script.js
```diff
@@ -0,0 +1,13 @@
...
+		var className = $toggle.attr('class');
+		var classBits = className.split('-checked');
+
+		if (classBits.length > 1) {
+			$toggle.attr('class', classBits.join(''));
+		} else {
+			$toggle.attr('class', className.replace('toggle', 'toggle-checked'));
+		}
+	});
+});
```
dominikwilkowski, line 13: Sorry about this but I think we should make the switch no-js.
Same way I did it with the form-checks.
Just use css.
We found that using css selectors here actually verifies that the html state is what it is rather than using js in react to control it. It would still be a controlled input but the visual layer of it does not depend on state.

components/switch/src/overrides/toggle.js
```diff
@@ -2,6 +2,7 @@
 
 import { jsx, useBrand, useMediaQuery } from '@westpac/core';
 import { sizeMap } from './_utils';
+import { getLabel } from '@westpac/core/src';
```
dominikwilkowski, line 5:
```suggestion
import { jsx, useBrand, useMediaQuery, getLabel } from '@westpac/core';
import { sizeMap } from './_utils';
```

components/table/src/overrides/caption.js
```diff
@@ -1,13 +1,14 @@
...
-import { jsx, useBrand } from '@westpac/core';
+import { jsx, useBrand, getLabel } from '@westpac/core';
 
 const Caption = ({ state, ...rest }) => <caption {...rest} />;
 
 const captionStyles = () => {
 	const { TYPE } = useBrand();
 
 	return {
+		label: getLabel('caption'),
```
dominikwilkowski, line 11: This should be table-caption
```suggestion
		label: getLabel('table-caption'),
```

components/table/src/overrides/tbody.js
```diff
@@ -1,10 +1,10 @@
 /** @jsx jsx */
 
-import { jsx } from '@westpac/core';
+import { jsx, getLabel } from '@westpac/core';
 
 const Tbody = ({ state, ...rest }) => <tbody {...rest} />;
 
-const tbodyStyles = () => ({});
+const tbodyStyles = () => ({ label: getLabel('tbody') });
```
dominikwilkowski, line 7: Are we not prefixing them? I'm not sure if we should now...

components/table/src/overrides/th.js
```diff
@@ -1,13 +1,14 @@
...
-import { jsx, useBrand } from '@westpac/core';
+import { jsx, useBrand, getLabel } from '@westpac/core';
 
 const Th = ({ state, ...rest }) => <th {...rest} />;
 
 const thStyles = (_, { bordered }) => {
 	const { COLORS } = useBrand();
 
 	return {
+		label: getLabel('th', { bordered }),
```
dominikwilkowski, line 11: Now I'm sure again... we should prefix them cause a cass like this `GEL-th` is not very helpful

components/text-input/blender/recipe.js
```diff
@@ -0,0 +1,228 @@
...
+export function AllStyles({ brand }) {
+	return (
+		<GEL brand={brand}>
+			<TextInput />
+			<TextInput size="small" placeholder="small" />
+			<TextInput size="medium" placeholder="medium" />
+			<TextInput size="large" placeholder="large" />
+			<TextInput size="xlarge" placeholder="xlarge" />
+			<TextInput invalid />
+			<TextInput inline />
```
dominikwilkowski, line 15: What ab out the combination of both `invalid` and `inline`?

dominikwilkowski: Oh also all the sizes in both of those combination?

components/text-input/src/overrides/textarea.js
```diff
@@ -19,6 +19,7 @@ const textareaStyles = (_, { size, width, inline, invalid, ...rest }) => {
 	)}`;
 
 	return mq({
+		label: getLabel('textarea', { size, width, inline, invalid }),
```
dominikwilkowski, line 22: Here I don't think we need to prefix...

components/well/blender/recipe.js
```diff
@@ -0,0 +1,35 @@
+import { GEL } from '@westpac/core';
+import React from 'react';
+
+import { Well } from '@westpac/well';
+
+export function AllStyles({ brand }) {
+	return (
+		<GEL brand={brand}>
+			<Well>Text</Well>
```
dominikwilkowski, line 9: Good call not to run nested Wells in here as they use css selectors but this may break in the future...
I would make sure what you document below is also in here at the least.

components/switch/blender/recipe.js
jaortiz: I already have it, its just further down in the examples

.changeset/three-needles-cheat.md
```diff
@@ -0,0 +1,6 @@
+---
+'@westpac/switch': major
+---
+
+- add blender support
```
jonnystening, line 5: We should use sentence case, dont ya reckon

components/switch/blender/recipe.js
```diff
@@ -0,0 +1,125 @@
+import { GEL } from '@westpac/core';
+import React from 'react';
+
+import { Switch } from '@westpac/switch';
+
+export function AllStyles({ brand }) {
+	return (
+		<GEL brand={brand}>
+			<Switch name="text" size="small" label="example" />
```
jonnystening, line 9: name must be unique

components/switch/src/overrides/toggle.js
```diff
@@ -36,24 +37,32 @@ const toggleStyles = (_, { size, checked }) => {
...
-			right: 0,
-			top: 0,
-			bottom: 0,
-			border: 'solid transparent',
-			borderWidth: borderWidthArr,
+		'input:checked ~ &': {
+			border: `1px solid ${COLORS.hero}`,
+			backgroundColor: COLORS.hero,
+
+			'&::after': {
```
jonnystening, line 52: pretty sure you don't need the & for these pseudo elements

components/symbol/blender/recipe.js
```diff
@@ -0,0 +1,74 @@
...
+		{
+			heading: 'A symbol with height 100',
+			component: () => (
+				<GEL brand={brand}>
+					<components.WBCLogo width={100} />
+				</GEL>
+			),
+		},
+		{
+			heading: 'A symbol with width and height of 100',
```
jonnystening, line 65: from memory i didn't it make it contain if both width and height are provided... think only the width is used

components/well/blender/recipe.js
```diff
@@ -0,0 +1,35 @@
...
+		{
+			heading: 'A well',
+			component: () => (
+				<GEL brand={brand}>
+					<Well>Text</Well>
+				</GEL>
+			),
+		},
+		{
+			heading: "Nested well's",
```
jonnystening, line 25: `heading: "Nested wells",`

PR #448, opened 2020-06-11
Popover blender support

dominikwilkowski:


components/popover/src/Popover.js
```diff
@@ -33,7 +34,10 @@ export const Popover = ({
 	} = useBrand();
 
 	const [open, setOpen] = useState(isOpen);
-	const [position, setPosition] = useState({ placement: 'top', empty: true });
+	const [position, setPosition] = useState({
+		placement: placement ? placement : 'top',
+		empty: !!!placement,
```
jaortiz, line 39: Isn't this just the same as a single `!` ?

dominikwilkowski: yeah but I like to make sure :D

.changeset/strong-mayflies-shake.md
```diff
@@ -0,0 +1,6 @@
+---
+'@westpac/popover': major
+---
+
+Added blender support and the `placement` prop
+You can now use `placement` to override the position detection for the popover panel.
```
jonnystening, line 6: please remove trailing `.` ;)

components/popover/blender/script.js
```diff
@@ -0,0 +1,22 @@
+function GELtoggelPopover($panel) {
+	var className = $panel.attr('class');
+	var classBits = className.split('-open');
```
jonnystening, line 3: haha love the name

components/popover/blender/script.js
```diff
@@ -0,0 +1,22 @@
...
+	}
+}
+
+$(function () {
+	$('[data-js="popover__version__"]').on('click', function () {
+		GELtoggelPopover($(this).next('[data-js="popover-panel__version__"]'));
+	});
+
+	$('[data-js="popover-closeBtn__version__"]').on('click', function () {
+		GELtoggelPopover($(this).parents('[data-js="popover-panel__version__"]'));
```
jonnystening, line 20: `$.parents()` ? 

I think `$.closest()` is the recommended way to get the matching ancestor

.changeset/strong-mayflies-shake.md
dominikwilkowski: Not a sentence?

components/popover/blender/script.js
dominikwilkowski: My jQuery is rusty at best... I figured `parents` would only look down the tree while `closest` looks both ways?

dominikwilkowski: `closest()` selects the first element that matches the selector, up from the DOM tree. Begins from the current element and travels up.

`parent()` selects one element up (single level up) the DOM tree.

`parents()` method is similar to parent() but selects all the matching elements up the DOM tree. Begins from the parent element and travels up.

So one could argue that `parents()` is more performant because it starts at the parent therefor has one less iterations 😬

jonnystening:
Re Blender HTML output of...

```<div class="GEL-popover-v1_0_0-beta_3">
	<button type="button" data-js="popover-1_0_0-beta_3" aria-expanded="false" class="GEL-button-v1_0_0-beta_3-look_hero-size_medium-Popover">
		<span class="GEL-button-v1_0_0-beta_3-text">
			The button text
		</span>
	</button>
	<div data-js="popover-panel-1_0_0-beta_3" class="GEL-popover-v1_0_0-beta_3-panel-placement_top">
		<h4 class="GEL-popover-v1_0_0-beta_3-heading">Your heading</h4>
		<div class="GEL-body-v1_0_0-beta_3-popover-body">
			The content of your popover
		</div>
		<button type="button" data-js="popover-closeBtn-1_0_0-beta_3" class="GEL-button-v1_0_0-beta_3-look_link-size_medium-popover-closebtn" aria-label="Close popover">
			<span aria-hidden="true" class="GEL-icon-v1_0_0-beta_3-color_inherit-size_small-button-icon-right">
				<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" role="img" focusable="false" class="GEL-icon-v1_0_0-beta_3-svg"><polygon fill="currentColor" fill-rule="evenodd" points="5.414 4 12 10.586 18.586 4 20 5.414 13.414 12 20 18.586 18.586 20 12 13.414 5.414 20 4 18.586 10.586 12 4 5.414"></polygon></svg>
			</span>
		</button>
	</div>
</div>```


Referring to... `<span aria-hidden="true" class="GEL-icon-v1_0_0-beta_3-color_inherit-size_small-button-icon-right">`
Wondering if we can refactor `<Icon />` so it doesn't need to render `color_inherit` in the class?

File: .changeset/strong-mayflies-shake.md (hunk 2 above)
Comment (jonnystening) [reply .changeset/strong-mayflies-shake.md:6]: Individual bullets don't get full stops, unless the bulleted item has 2+ sentences

- I am a thing
- I am two things. See how I have two full stops.

File: components/popover/blender/script.js (hunk 4 above)
Comment (jonnystening) [reply components/popover/blender/script.js:20]: Up to you. I was under the impression that if there were additional matching elements up the tree it would select all, rather than the one as we'd expect.

File: .changeset/strong-mayflies-shake.md (hunk 2 above)
Comment (dominikwilkowski) [reply .changeset/strong-mayflies-shake.md:6]: I don't see bullet points though? I guess the first sentence is missing a period?

**Comment by dominikwilkowski** [comment]
This is great! Good catch. Will have a look now

**Comment by dominikwilkowski** [comment]
I just had a look and this is a button concern. Buttons with icons generate this html:
```
<button type="button" class="GEL-button-v1_0_1-look_primary-size_medium">
	<span class="GEL-button-v1_0_1-text">
		Your primary button text
	</span>
	<span aria-hidden="true" class="GEL-icon-v1_0_1-color_inherit-size_small-button-icon-hasChildren-right">
		<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" role="img" focusable="false" class="GEL-icon-v1_0_1-svg"><polygon fill="currentColor" fill-rule="evenodd" points="12 0 0 12 4 12 4 24 10 24 10 17 14 17 14 24 20 24 20 12 24 12"></polygon></svg>
	</span>
</button>
```

This is because the default color for icons is `muted` and `inherit` is a mutation which is why it's called out.

jonnystening:
Yeah that’s right. Is there another way to do it?

dominikwilkowski:
I don't see how.. But maybe that's an issue we can log for buttons and not have it in here?

.changeset/strong-mayflies-shake.md
jonnystening: Sorry thought the individual lines got turned into bullets in the resulting changelog file

dominikwilkowski: Oh do they? I don't know! :D

jonnystening: Don't think so. Mind adding a training `.` after the word 'prop' (line5)

PR #451, opened 2020-06-12
Input group blender support

dominikwilkowski:
Ready to be reviewed


components/input-group/src/overrides/select.js
```diff
@@ -1,13 +1,14 @@
...
-import { jsx } from '@westpac/core';
+import { jsx, getLabel } from '@westpac/core';
 import { Select as SelectInput } from '@westpac/text-input';
 
 const Select = ({ state: { size, data }, ...rest }) => (
 	<SelectInput size={size} data={data} {...rest} />
 );
 
-const selectStyles = (_, { position }) => ({
+const selectStyles = (_, { position, size }) => ({
```
dominikwilkowski, line 10: This was interesting!
I had to include `size` even though it was not used in the styles because it's used in `SelectInput` in the component and this label overrides the label from the `SelectInput`.

jonnystening: ^ should we make note of that in the code with a comment?

jonnystening:
This still a WIP @dominikwilkowski?

dominikwilkowski:
Yes, I have to finish the docs... happy for you to take over if you like :)

dominikwilkowski:
Finished with docs now. Good to be reviewed @jonnystening @jaortiz

PR #456, opened 2020-06-16
WIP: Modal blender support

jaortiz:
@dominikwilkowski Does the blender support portal components? Nothing is being generated currently.


dominikwilkowski:
For portals we will have to use something like https://github.com/MichalZalecki/react-portal-universal.
Think SSR.

PR #460, opened 2020-06-18
WIP: Progress rope blender support

jaortiz:
@dominikwilkowski I've fixed some things in progress rope but it's still not working correctly. I did some exploring and it seems like the correct state is not being used when rendering out the component in the blender. It looks like it uses the first instance of a render before any hooks have run to update the component to the correct state. Any ideas?

### DIFF
--- a/components/progress-rope/blender/recipe.js
+++ b/components/progress-rope/blender/recipe.js
@@ -1,0 +1,184 @@
+import { GEL } from '@westpac/core';
+import React from 'react';
+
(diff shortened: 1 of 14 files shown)


dominikwilkowski:
@jaortiz The blender simply uses [renderToStaticMarkup](https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup). So `useEffect` won't execute at all. Maybe this helps?

PR #461, opened 2020-06-26
Grid blender support

jonnystening:
GEL3.0 implementation of the Bootstrap Grid, integrating with GEL3.0 `LAYOUT` (breakpoints) and `SPACING` tokens.

I have implemented this as cleanly as possible via a simple `/components/grid/blender/bootstrap/bootstrap-grid.js` entry point imported into the grid recipe. I pass required config values (breakpoints and spacing) through build functions in this file, rather than having multiple connection points. The structure of the JS files matches Bootstrap's SCSS structure which will help with maintainability. 

The solution differs from Bootstrap Grid's CSS output in the following instances...

- GEL3.0 uses a fluid container by default
  - GEL3.0 Blender Grid `.container` is BS `.container-fluid`
  - GEL3.0 Blender Grid `container-fixed` is BS `.container`
  - BS responsive containers `.container-{bp}` are not implemented
- Different default gutter values
  - BS `.row` gutters are `1.5rem` across all breakpoints by default, GEL3.0 uses responsive array values `['12px', '24px']`
- Different spacing values
  - BS gutter `.g-{x}` and padding `.p-{x}` utility classes: in BS `x` is 1–5, in GEL3.0 `x` is 1–10

I've brought through all the example windows from the Bootstrap Grid docs. Some changes were made to the Blender code in order to provide heading/subheading output and some additional example window styling options in the docs, so make sure you pull and build the latest Blender code.

[Sample docs output](https://github.com/WestpacGEL/GEL/files/4849062/screencapture-file-Users-m044451-Downloads-blender-docs-packages-grid-html-2020-06-30-12_21_50.pdf)


components/grid/blender/_utils.js
```diff
@@ -0,0 +1,6 @@
+export const rowMap = {
+	columns: 12,
+	gap: [12, 24],
+};
+
+// export const colMap = {};
```
dominikwilkowski, line 6: Can clean this...

jonnystening: Yeah sweet no probs

Could also get the `columns` and `gap` values above by exporting the defaultValues out of `Grid.js`. Is that worth considering?

jonnystening: ^ like this https://github.com/WestpacGEL/GEL/commit/ba2682bbcd35b2c9dc7770fb6985732fe3e8a335 ?

jonnystening:
@dominikwilkowski @jaortiz I've updated the Blender Grid docs to explain the connection with Bootstrap, and points of difference. Let me know what you think...

Screen Shot 2020-07-03 at 8 27 43 am

PR #467, opened 2020-07-06
fixed icon state and showed how to override an icon entirely

dominikwilkowski:
The icon omited the `icon` name in it's state. Must have been deleted when we moved to nested `state` prop.
I also showed how one may override the entire icon paths if they chose to.

### DIFF
--- a/components/icon/examples/60-overrides.js
+++ b/components/icon/examples/60-overrides.js
@@ -3,6 +3,7 @@
 import { GEL, jsx } from '@westpac/core';
 import {
 	AddIcon,
+	BankIcon,
 	CalendarIcon,
 	DeleteIcon,
 	GridIcon,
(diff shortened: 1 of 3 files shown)


components/icon/examples/60-overrides.js
```diff
@@ -15,7 +16,7 @@ import { Fragment } from 'react';
 
 import { Intopia } from '../../../helpers/example/components/Intopia.js';
 
-const Wrapper = ({ children, icon, color, size, assistiveText, ...rest }) => (
+const Wrapper = ({ children, state: { icon }, ...rest }) => (
```
dominikwilkowski, line 19: first this was wrong which caused the overrides page to fail on the names

components/icon/src/Icon.js
```diff
@@ -34,6 +34,7 @@ export const Icon = ({
 		color,
 		size,
 		assistiveText,
+		icon,
```
dominikwilkowski, line 37: second this was missing from the state

jonnystening:
Nice thanks @dominikwilkowski 

LGTM

PR #498, opened 2020-08-17
Better Core specificity

jonnystening:
This solution should mean we can remove the `&&` stacked (double) class solution we’re currently using to increase specificity so our component styling takes precedence over reset styling (mostly affecting h1, buttons, form inputs etc).

The stacked class solution produced classes (such as `.css-1f19ls7-heading-size_1.css-1f19ls7-heading-size_1 {}`) in order to increase specificity.

This PR solves the specificity issue differently by adding a root parent selector to our styling to increase specificity.

**How it’s implemented...**
- Adds a `GEL` className on the `<GEL />` wrapper component (e.g. `<div class="GEL css-kr80hg-Core" />`)
- Prepends all css selectors generated from child components of `<GEL />` with `.GEL` parent selector (e.g. `.GEL .css-17eooxh-heading-size_1` {})
- Core won’t prepend `.GEL` css selector for itself or any descendent selector styles. e.g.:
  - `<GEL />` nested inside a `<GEL />` won’t generate a `.GEL .css-kr80hg-Core {}` selector (it will simply output `.css-kr80hg-Core {}`), and 
  - Reset styles in Core such as `h1 {}` won't output as `.GEL .css-kr80hg-Core h1 {}` (it will simply output as `.css-kr80hg-Core h1 {}`)

**Has been tested with...**
 - Nested `<GEL />` components (e.g. `<GEL><GEL><Heading>This is a heading</Heading></GEL></GEL>`) 
 - Selectors using the 'Current Class' symbol `&` (e.g. stacked and combinator selectors (e.g. &&, & > & etc)
 - `@font-face` declarations
 - `@media` queries using Emotion (facepaint) array values
 - Child selectors, by class and tag (e.g. `css={{ 'div &' : {} }}` and `css={{ 'div p': {} }}`)

New examples test page added to Core https://github.com/WestpacGEL/GEL/blob/feature/core-specificity/components/core/examples/04-CSS-selectors.js

**Pros**
- Means we can rely on css to work as intended, without needing stacked class selectors (`css={{ '&&': {} }}`) and !important values
- Makes overrides easier to implement, Devs don't need to think about underlying style specificity
- Makes building custom components easier, Devs don't need to worry about GEL reset styling overriding their own styling

**Cons**
- Slightly longer CSS selectors 🤷‍♂️
- Reliance on an emotion/stylis plugin feature(?)
- Requires @emotion/cache


components/core/src/reset.js
```diff
@@ -2,15 +2,18 @@
...
+/* Note: The following styles were not included as they apply to the document specifically...
+html {
+	line-height: 1.15;
+	-webkit-text-size-adjust: 100%;
+}
+body {
+	margin: 0;
+} 
+*/
+// TODO: consider moving to <Global />, but note will apply to document (outside of <GEL /> wrapper)
```
dominikwilkowski, line 15: This would break things what AEM wanted with multiple React apps in the same page

components/heading/examples/00-Heading.js
```diff
@@ -1,138 +1,141 @@
...
-			<Heading size={2}>Heading size: 2</Heading>
-			<Heading size={3}>Heading size: 3</Heading>
-			<Heading size={4}>Heading size: 4</Heading>
-			<Heading size={5}>Heading size: 5</Heading>
-			<Heading size={6}>Heading size: 6</Heading>
-			<Heading size={7}>Heading size: 7</Heading>
-			<Heading size={8}>Heading size: 8</Heading>
-			<Heading size={9}>Heading size: 9</Heading>
-			<Heading size={10}>Heading size: 10</Heading>
+		<Fragment>
```
dominikwilkowski, line 9: It's a long diff but I can't see why there is a `Fragment` used...?

helpers/example/components/AppDocs.js
```diff
@@ -11,44 +11,50 @@ import { HomeDocs } from './HomeDocs';
...
-	const [brand, setBrand] = useState('WBC');
+	const [brandName, setBrandName] = useState('WBC');
+
+	const brand = BRANDS[brandName];
 
 	return (
 		<Router>
 			<StrictMode>
-				<GEL brand={BRANDS[brand]}>
+				<Global
```
dominikwilkowski, line 21: Do all consumers now have to do this?

website/src/components/layout/layout.js
```diff
@@ -57,7 +57,7 @@ const Wrapper = (props) => {
 		return (
 			<GEL brand={brandOverrides(brands['WBC'])}>
 				<Global styles={gelFonts} />
-				<Normalize />
+				{/* <Normalize /> */}
```
dominikwilkowski, line 60: Can likely be deleted?

components/core/src/reset.js
jonnystening: The TODO comment was referring to just moving the two styles removed from the reset up into `<Global />`. Just these ones...

```
html {
	line-height: 1.15;
	-webkit-text-size-adjust: 100%;
}
body {
	margin: 0;
} 
```

They do nothing when rendered inside our wrapper. It would attempt to at least set the document styling to a reliable starting point. The `webkit-text-size-adjust` is important, margin reset too I guess. But yes moving these to Global space could possibly cause issues for existing pages, only minimal 'issues' though... what do you reckon @dominikwilkowski ?

components/heading/examples/00-Heading.js
jonnystening: Was left over from something I was playing with. Now removed.

helpers/example/components/AppDocs.js
jonnystening: Yeah these are just styles we want for our docs right?

The `body { margin: 0 }` reset style is mentioned in TODO conversation above. It's in our core reset already, but currently not used.

website/src/components/layout/layout.js
jonnystening: I think so yes... not sure why they were there. We have the exact reset styling in Core already

helpers/example/components/AppDocs.js
jonnystening: Oh do you mean, do all consumers have to keep <Global /> outside of GEL? 

Our stylis plugin function in Core was prepending Global styles with `.GEL` (which we don't want). I may have just found a way to fix that, which will mean Global styles are ignored.

components/core/src/reset.js
dominikwilkowski: I would not touch things outside of your system. You don't know what the conditions are...

jonnystening:
May have just hit a snag with this...

`& + &` selectors render as `.GEL .css-hash + .GEL .css-hash` 🤦‍♂️

jonnystening:
> May have just hit a snag with this...
> 
> `& + & {}` selectors output as `.GEL .css-hash + .GEL .css-hash {}` 🤦‍♂️

Have found a solution... can filter these out with a regex

helpers/example/components/AppDocs.js
jonnystening: That is now implemented. Styles passed to `<Global />` will not be processed (not prepended with the `.GEL ` parent class).

jonnystening:
New approach. Much simpler. Now prepending the style declaration during a later stage in stylis’ processing. 

- `.GEL ` is added to the very beginning of the style string, if it's not a @media query
- `.GEL ` is added within the style string if it's within a @media query block

This approach mitigates all the internal `.GEL ` double ups I was having to weed out with the earlier approach.

components/a11y/examples/20-overrides.js
```diff
@@ -4,7 +4,7 @@ import { GEL, jsx, useBrand } from '@westpac/core';
 import { HouseIcon } from '@westpac/icon';
 import { SkipLink, VisuallyHidden } from '@westpac/a11y';
 
-const SkipLinkWrapper = ({ children, ...rest }) => (
+const SkipLinkOverride = ({ state, children, ...rest }) => (
```
dominikwilkowski, line 7:
```suggestion
const SkipLinkOverride = ({ state: _, children, ...rest }) => (
```

components/core/package.json
```diff
@@ -53,6 +53,7 @@
 	},
 	"dependencies": {
 		"@babel/runtime": "^7.9.6",
+		"@emotion/cache": "^10.0.29",
```
dominikwilkowski, line 56:
```suggestion
		"@emotion/cache": "^10.0.28",
```

components/core/src/Core.js
```diff
@@ -1,37 +1,82 @@
...
 
-import { jsx } from '@emotion/core';
+import { jsx, CacheProvider } from '@emotion/core';
+import { Fragment, useState } from 'react';
+import createCache from '@emotion/cache';
 
 import { useBrand } from './Brand';
 import { reset } from './reset';
 
+const label = 'Core';
```
dominikwilkowski, line 10:
```suggestion
const label = 'Core'; // for use in stylis plugin and the css label
```

dominikwilkowski (commented):
Looks good, just some small tweaks

PR #501, opened 2020-08-19
Fix website example block

jonnystening:
We were using our `<Well />` component to render the example blocks, but since `<Well />` renders as a `<Body />`, and `<Body />` styles links etc, examples such as `<Breadcrumb />` were being affected by Body styles.

It's best show our examples accurately in their native environment, not affected by being rendered inside other component(s).


jaortiz (approved):
LGTM

PR #508, opened 2020-09-01
Update wbc logos

jonnystening:
WBC logos and WBC Red avatar logos updated

Remove WBCLiveLogo, no longer required

Awaiting decision on Westpac Live logo ('live' script text gradient colours)


jaortiz (approved):
LGTM

PR #510, opened 2020-09-03
Tidy up destructured props

jonnystening:
- rename destructured state props as _ and ensure prop order is consistent
- rename component override examples to be consistent


jaortiz (approved):
LGTM

PR #514, opened 2020-09-07
Multi-brand logos

jonnystening:
Large and Small multi-brand logos with `align` options


jaortiz (approved):
LGTM

jonnystening:
Will add a commit to use these on the website before merging

PR #516, opened 2020-09-07
label blender re-architecture

jaortiz:


jaortiz:
Example output
```
<span class="GEL-label-v1_0_0-beta_2-hero GEL-label-v1_0_0-beta_2">
	Your hero label text
</span>
```

PR #528, opened 2020-09-14
Optimise website brand switcher logos

jonnystening:
Individually optimises brand switcher logo sizing/alignment (since we're not using small multibrand logos)


jaortiz (approved):
LGTM

PR #534, opened 2020-09-18
removed the stylis plugins as it has moved to the blender

dominikwilkowski:
We have moved the stylis plugin into the blender via https://github.com/WestpacGEL/blender/pull/37
We need to remove it from here now.

### DIFF
--- a/components/core/src/Core.js
+++ b/components/core/src/Core.js
@@ -1,53 +1,9 @@
 /** @jsx jsx */
 
-import { jsx, CacheProvider } from '@emotion/core';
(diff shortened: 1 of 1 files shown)


jaortiz:
@dominikwilkowski Why are we removing it from here? We still need it here to elevate component specificity so the reset doesn't override them. Shouldn't we just add a prop to this component to turn it off when we use it in the blender?

dominikwilkowski:
Sure we can do that but I would probably try to run this at runtime. Up to you folks.

PR #574, opened 2020-11-10
Update modal (react and blender)

jonnystening:
- [x] React modal
- [x] Blender modal

Still has a few little bugs, but we can come back to these...
- Focus trap in react isn't yet working (waiting for react-focus-on to be updated)
- Focus trapping in blender modal isn't perfect (Shift + Tab) gets you out of the modal context


jaortiz (approved):
LGTM

PR #576, opened 2020-11-12
Fix keyCode

jonnystening:
event.keyCode is being deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode


jaortiz (approved):
LGTM

PR #605, opened 2020-12-15
Fix website body styling

jonnystening:
Strips tabcordion PanelBody styling, removes use of <Body /> component and overrides tabButton styling

Fixes:
- broken breadcrumb styling in example box
- incorrect page padding
- tabButton hover styling


jaortiz (approved):
LGTM

PR #608, opened 2020-12-17
New testing brand selector UI

jonnystening:
Replaces the horizontal button UI with a more scalable Select based control


jaortiz (approved):
LGTM

PR #618, opened 2021-01-19
Fix pagination list role

jonnystening:
Add list role to enhance UX in VoiceOver

Now consistent with other lists... breadcrumb, progress-rope etc


jaortiz (approved):
LGTM

PR #622, opened 2021-02-02
Dom/page tabs

dominikwilkowski:
Closing https://trello.com/c/kZtNDkSb/110-dom-website-routing

This will enable the navigation of the in-page tabs without them being handled as a page in the browser.
Mitchell helped me with fixing a bug where each page reload would also re-instantiate the apollo client which would make the site much slower than it should be. It also helps with using the router of next better without having it refresh the entire page.

This PR includes also a fix to tabcordion itself where `instanceIdPrefix` would cause a re-render even when it's not used and a fix to the pesky react error complaining about the missing `key` to an iteratable react component.

### DIFF
--- a/components/tabcordion/examples/50-callbacks.js
+++ b/components/tabcordion/examples/50-callbacks.js
@@ -77,6 +77,7 @@ function Example({ brand }) {
 			<hr />
 			<Tabcordion
 				mode="accordion"
+				openTab={1}
 				onOpen={onOpen}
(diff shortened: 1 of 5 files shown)


jaortiz (approved):
LGTM

dominikwilkowski:
You might want to include a changeset for the tabcordion actually.

jonnystening (approved):
Mind adding the changeset plz @jaortiz

PR #623, opened 2021-02-03
Added documentation for moving data from staging to live and reverse

dominikwilkowski:
…and made the process a bit more transparent


jonnystening (approved):
Sweet moves

jaortiz (approved):

PR #624, opened 2021-02-03
New blender JS public API pattern rollout

jonnystening:
Updates to existing components with blender script functionality. Now consistent with new Tabcordion JS pattern.


jaortiz (approved):
LGTM

PR #625, opened 2021-02-04
Remove button-group faint look option

jonnystening:
This was never intended to be an option. It was never communicated. 

Issue: It's not clear enough when a button is selected, and is not accessible.


jaortiz (approved):
LGTM

PR #647, opened 2021-02-24
Dom/paper cuts

dominikwilkowski:
Just some small things I wanted to fix while working on the site.

- Fixed react key warning
- Fixed emotion keys


jonnystening (approved):
Awesome, was so sure that WebkitFontSmoothing wasnt working without the -, hence why we were having to put up with the annoying error. Just checked it, all working. I must have been wrong. Thanks Dom 👏

jaortiz (approved):

PR #649, opened 2021-02-24
added documentation links back and pinned core to the top of our download list

dominikwilkowski:
fixes https://trello.com/c/vypyWSxi/277-dom-downloads-page-re-instate-correct-documentation-link
fixes https://trello.com/c/5UEjIRXw/413-dom-add-dummy-core-selection-in-downloads-page

### DIFF
--- a/website/src/pages/design-system/downloads/index.js
+++ b/website/src/pages/design-system/downloads/index.js
@@ -16,7 +16,10 @@ import { Textarea, Select } from '@westpac/text-input';
 import { Container, Grid, Cell } from '@westpac/grid';
 import { Button } from '@westpac/button';
 import { useTransition, animated } from 'react-spring';
+import { useQuery } from '@apollo/react-hooks';
 import merge from 'lodash.merge';
+import gql from 'graphql-tag';
+import Link from 'next/link';
 
 import { Section, SectionHeading } from '../../../components/section';
 import { Body } from '../../../components/body';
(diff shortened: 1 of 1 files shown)


jonnystening:
@dominikwilkowski the docs links are missing `/design-system` in the URL

dominikwilkowski:
Good spot. Fixed the `BASE_URL` bug. Not sure about `as` and `href` with the old packages of next. I'd keep it that way until we update to new packages

jonnystening (approved):

jaortiz (approved):

jonnystening:
> ...Not sure about `as` and `href` with the old packages of next. I'd keep it that way until we update to new packages

All good 

```
href - The path or URL to navigate to. This is the only required prop
as - Optional decorator for the path that will be shown in the browser URL bar. Before Next.js 9.5.3 this was used for dynamic routes, check our previous docs to see how it worked
```
https://nextjs.org/docs/api-reference/next/link

dominikwilkowski:
Note that this link refers to the latest version of `next/link`... We are not on the latest version by a long shot and there were changes to the `as` prop I remember...

jonnystening:
> Note that this link refers to the latest version of `next/link`... We are not on the latest version by a long shot and there were changes to the `as` prop I remember... We are on
> 
> ```
> next@^9.1.1, next@^9.3.2:
>   version "9.3.6"
> ```

I removed `as` and it worked (at least until my machine blew up and I got blocked by rollup replace plugin)

dominikwilkowski:
Yes because it would load the page entirely and not use a dynamic route...

PR #673, opened 2021-03-31
fixed page header, fixed rerender on scroll

dominikwilkowski:
The header was re-rendering due to the scroll events added all over the place. I consolidated most of them and kept the footer and homepage once (for another day)

This now resolves https://trello.com/c/FCO631qQ/316-svg-filter-in-docs-search-clears-on-scroll

I also noticed that the header was really "badly" done with no transition phases etc. So I fixed that as well

And lastly I found an issue in the font loader that I fixed.


jaortiz (approved):
LGTM

jonnystening:
You're not reusing the `<Heading /> responsive `size` feature now tho...

dominikwilkowski:
Well I am but not via the prop @jonnystening

jonnystening:
@dominikwilkowski `body.hasScrolled` was not being added to the downloads page (a hard-coded page) so the transition wasn't being applied to that page.

Using this approach to share the logic... you ok with that? https://github.com/WestpacGEL/GEL/pull/673/commits/183d307a87072e39e57131a39ab43b5cb5d89f81

dominikwilkowski:
Yeah looks fine @jonnystening 
We might want to add this to the `_app.js` file so that it's applied universally to all pages rather than adding it into a hook?

jonnystening:
> Yeah looks fine @jonnystening
> We might want to add this to the `_app.js` file so that it's applied universally to all pages rather than adding it into a hook?

An den?... https://github.com/WestpacGEL/GEL/pull/673/commits/70e798cdc252dc0b44ab9cc89713b72b6607a2a8

website/src/components/fontPreloader/FontPreloader.js
```diff
@@ -224,10 +224,9 @@ export const FontPreloader = ({ title = '', ...rest }) => {
 					left: '10px',
 					width: '1px',
 					height: '1px',
-					padding: '10px',
 					overflow: 'hidden',
 					zIndex: 99999,
-					background: '#fff',
+					background: 'transparent',
 				}}
```
jonnystening, line 231: Hey @dominikwilkowski can't we treat this like our VisuallyHidden component and use that component's styling?

Also visibility isn't a valid attribute... https://github.com/WestpacGEL/GEL/commit/c756e2fb877001e1e63e4568cca4ca8180a20590#r49242571

jonnystening (approved):

PR #684, opened 2021-04-20
Normalize / specificity solution

jonnystening:
This PR removes the custom .GEL parent class (higher specificity) solution in order to illustrate the issue.

# Example
A simple example to illustrate... styling of our link button `<Button href="#">Link button example</Button>`

```
<!-- GEL wrapper-->
<div class="css-o867iy-Core"> 
   <a href="#" class="css-n69q6e-button">Link button example</a>
</div>
```

### Without a solution
With common normalize resets like `a { background-color: transparent; }` being applied to the scoped GEL wrapper we end up with...
```
/* Reset */
.css-o867iy-Core a {
    background-color: transparent;
}
```

But then Emotion styles our button component as...
```
.css-n69q6e-button {
    background-color: #1F1C4F;
    ...
}
```
...we end up with the reset styling having higher specificity than our own button's specificity. The reset wins and we get no backgrounds showing in link buttons.

e.g.
Screen Shot 2021-04-20 at 1 18 44 pm

## 1. The GEL parent class solution
This approach was an attempt to increase the specificity of our component's styling, so that they always had higher specificity than the global reset styling.

The GEL parent class solution solved this by prepending a .GEL parent class...
```
/* Reset */
.css-o867iy-Core a {
    background-color: transparent;
}
```
```
.GEL .css-n69q6e-button {
    background-color: #1F1C4F;
    ...
}
```
...the GEL parent class increased our button's specificity to be equal to the reset's (taking precedence because specified further down the stylesheet)

e.g.
Screen Shot 2021-04-20 at 1 19 20 pm

## 2. Global reset solution
When the reset styling was applied globally to the document, we'd end up with this...
```
/* Reset */
a {
    background-color: transparent;
}
```
```
.css-n69q6e-button {
    background-color: #1F1C4F;
    ...
}
```
...and the our button component would have equal specificity and win (as with the example above).

e.g.
Screen Shot 2021-04-20 at 1 19 20 pm

## 3. Mix it in solution
Pull the Normalize styling apart and apply to components as needed. We could make the entire Normalize library an opt-in feature applied globally via a prop on the <GEL /> wrapper.


jonnystening:
I propose we just promote the normalize resets to the document level, so they're no longer scoped to the GEL wrapper. 

Most if not all of these resets will likely already exist in websites/applications and we do offer a `noReset` boolean prop that devs can use if they don't want to have these reset styles set on the document.

We can still leave our standard Core resets scoped to the GEL wrapper...
```
.css-1au7r59-core {
    line-height: x;
    color: x;
    font-family: x;
    font-weight: x;
    etc
}
```

We'd only be moving the normalize styling back up to global scope...

```
main {
    display: block;
}
h1 {
    font-size: 2em;
    margin: 0.67em 0;
}
hr {
    box-sizing: content-box;
    height: 0;
    overflow: visible;
}
pre {
    font-family: monospace, monospace;
    font-size: 1em;
}
a {
    background-color: transparent;
}
etc
```

This would make all my hurt go away 🙌

jonnystening:
Solution we're going with is...

Normalize CSS styling will be split up and applied to components as required

Normalize will be made an optional opt-in via prop on the GEL wrapper. It will be applied at the global document scope if enabled.

jaortiz (approved):
LGTM

PR #686, opened 2021-04-21
Progress-bar: make visual progress text optional

jonnystening:
New `noLabel` boolean prop. Name ok? Happy with approach?

Just waiting to hear if there are further changes required to the component. Possibly the ability to use units other than %.


jaortiz (approved):
LGTM

jonnystening:
Ability to use units other than % coming in a separate PR

PR #694, opened 2021-05-03
Fix website page header

jonnystening:
Fixes page header scroll shadow missing on component and download pages

also fixes downloads page blender options list styling


jaortiz (approved):
LGTM

PR #695, opened 2021-05-03
Field, Fieldset and Autocomplete

jaortiz:


components/field/src/Field.js
```diff
@@ -0,0 +1,154 @@
...
+			</Label>
+			{hint && (
+				<Hint state={state} {...hintAttributes(state)} css={hintStyles(state)}>
+					{hint}
+				</Hint>
+			)}
+			{error && (
+				<Error message={error} state={state} {...errorAttributes(state)} css={errorStyles(state)} />
+			)}
+			{children(inputProps)}
```
dominikwilkowski, line 93: You can also do this via mapping over the Children so that the composition isn't impacted. We did this in the List component etc. Might be best to stick to the same thing?

dominikwilkowski (commented):
The only thing I found that was weird is the way to add props to the children to the `Field` component. Otherwise it all looks good.

components/field/src/Field.js
jaortiz: @dominikwilkowski not sure what you mean here. Im passing the input props to the children so they become available as render props.

dominikwilkowski: All you need is making sure the children have the `inputProps` available to them right? That's the same problem in `List` where we need `nested` in each child. The difference to your solution is that children have to be written as functions now (render props) and not composed as normal children. The change is not necessary and might be confusing to our users.

dominikwilkowski (approved):
LGTM

PR #705, opened 2021-05-17
Feature/repeater

jaortiz:


dominikwilkowski (commented):
I like it. Looks great. I would have used Children as `component` so instead of:
```jsx
<Repeater component={Repeat} />
```

I would do this for composition:

```jsx
<Repeater>
  <Repeat />
</Repeater>
```

And that would allow you do a lot more as well no?

dominikwilkowski (approved):
LGTM

PR #710, opened 2021-05-24
Selector: Add secondary label ellipsis truncation

jonnystening:
Also added a few examples for bank account numbers as hints


jaortiz (approved):
LGTM

PR #721, opened 2021-06-01
fixed tabcordion

dominikwilkowski:


components/tabcordion/src/overrides/panelBody.js
```diff
@@ -31,7 +31,7 @@ const PanelBody = ({ state: { mode, selected }, setPanelHeight, ...rest }) => {
 	});
 	const AnimatedBody = animated(Body);
 
-	return <AnimatedBody ref={measureRef} style={fade} {...rest} />;
+	return <Body ref={measureRef} {...rest} />;
 };
```
jonnystening, line 35: Can we put `<AnimatedBody />` back?... you've removed the fade effect 😭

jonnystening (changes requested):

components/tabcordion/src/overrides/panelBody.js
dominikwilkowski: ooops

jonnystening (approved):
LGTM

PR #726, opened 2021-06-04
Fix button-dropdown

jonnystening:
Fixes issue with useOutsideClick hook in react v17 (button-dropdowns were playing up in the website)

- Updated hook to match Joss' version found in website/field-content
- Updated button-dropdown to use new hook api
- Add some further hooks we can use elsewhere


jaortiz (approved):
LGTM

PR #733, opened 2021-06-14
Dom/svg download

dominikwilkowski:
Added download buttons to each SVG assets dynamic blocks and the API route for the generation of the SVGs.

I have not tested the nginx config... so let's do that in staging.


jonnystening (approved):
Amazing work @dominikwilkowski, especially over your long weekend ⭐️

PR #738, opened 2021-06-17
Form-check: New <FormCheckReveal> component

jonnystening:
I'm not sold on the prop name of `show`, need to think


jaortiz (approved):
LGTM

PR #751, opened 2021-07-02
Selector: tweak type with next indicator

jonnystening:
Previous attempt didn't work for keyboard users; who select the first option on arrival, and only fire onChange on spacebar press or first arrow down (so wouldn't be able to actually choose anything other than the first or second option, before being taken to the next step/page in the form) 😬

New approach has been implemented as `type="button"` rather than a styling prop (`nextIndicator`). It uses buttons with `aria-pressed` attributes to indicate (a11y) state change. Rather than a separate input for each option, there is one hidden input with dynamic value. 

The button elements have a `data-value` attribute for use in the blender. Some jQuery will take the currently selected button onClick, and set the hidden input value to the button's data-value.

### DIFF
--- a/components/selector/examples/00-examples.js
+++ b/components/selector/examples/00-examples.js
@@ -23,15 +23,15 @@ function Example({ brand }) {
 
 			<h2>Types</h2>
 
-			<h3>Radio</h3>
-			<Selector type="radio" name="example-radio">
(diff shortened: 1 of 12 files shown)


jaortiz (approved):
LGTM

Issue #772, opened 2021-08-05
Consider adding fontDisplay option to useFonts ot using `font-display: swap` by default


wojtekmaj:
**Is your feature request related to a problem? Please describe.**

To avoid [FOIT](https://css-tricks.com/fout-foit-foft/), `font-display: swap` can be used so that a fallback font is used while the main font is loading.

**Describe the solution you'd like**

Non-breaking solution:

*
    ```js
    useFonts({ path: '/some/path', fontDisplay: 'swap' }); 
      // Can be 'auto', 'block', 'swap', 'fallback', 'optional'
    ```

* Leverage [`size-adjust`, `ascent-override`, `descent-override`, `line-gap-override`](https://web.dev/css-size-adjust/) to prevent cumulative layout shift (this can be done independently)

Breaking solution:

* Add `font-display: swap` by default. 
* Leverage [`size-adjust`, `ascent-override`, `descent-override`, `line-gap-override`](https://web.dev/css-size-adjust/) to prevent cumulative layout shift (this can be done independently)

**Describe alternatives you've considered**

```js
const fonts = useFonts({ path: '/static/fonts/' });
fonts[''].forEach((font) => { font['@font-face'].fontDisplay = 'swap'; });
```

🤢

dominikwilkowski:
I totally agree with your sentiment and we had a long internal discussion about this. Adding an option would be a good idea actually. Will see when I get time next.

Can I ask where you're using this lib? Would be good to understand more about your use case

wojtekmaj:
@dominikwilkowski Not sure if I can share that publicly, so I dropped you a DM on Twitter.

jonnystening:
@wojtekmaj feel free to contact us directly if you need anything gel@westpac.com.au or via Teams

wojtekmaj:
Sorry, Teams? How do I join? I see only Slack in your footer, but it doesn't seem to be public. Would love to chat with you guys from time to time.

dominikwilkowski:
This was merged btw: #773
Closing this. I'm sure it will be published soon

PR #773, opened 2021-08-06
added a way for someone to override anything inside the useFont hook

dominikwilkowski:
Looking into #772 this could be a neat solution


jonnystening (approved):
LGTM

PR #780, opened 2021-08-09
Fix mouse mode never turned back on, refactor isMouseMode to focus-visible polyfill-like

### LINKED ISSUE #779: Focus outline appears during mouse navigation after the user hits Tab once
**Describe the bug**

Focus outline is programmed to be only shown when the user uses keyboard navigation. That works, until the user hits Tab key once. Once done, there's no going back: The user is "stuck" in keyboard mode forever. Well, until refresh.

**To Reproduce**

1. Click a button. 
image
Nice.
2. Hit Tab. Next button focuses with an outline.
image
Nice.
3. **Click** the first button again. Outline appears despite the user using mouse.
image

**Expected behavior**
Focus outline not to be shown for mouse interactions after using keyboard navigation.

**Screenshots**
See above.

**Desktop (please complete the following information):**

- OS: [e.g. iOS] macOS Monterey
- Browser [e.g. chrome, safari] Chrome
- Version [e.g. 22] 92

**Additional context**
Looks like the problem lies here:
https://github.com/WestpacGEL/GEL/blob/55b880f4edde06a04355bd9cf075b5dffd2d2cbe/components/core/src/useFocus.js

Depending on the list of browsers you'd like to support, I'd recommend:
* Opting for native `focus-visible`
* Using [WICG's focus-visible polyfill](https://github.com/WICG/focus-visible) which does a little more advanced heuristics on when the outlines should appear than `useFocus`.

BTW I may not understand `useFocus` fully but I fail to understand why are we adding inline scripts instead of simply firing them by using useEffect. Not only this pollutes the DOM, but also includes script body as a string which will not be nicely minified by terser and other minifiers.

wojtekmaj:
Fixes #779.

https://user-images.githubusercontent.com/5426427/128701591-10b29901-f49a-4ab3-923f-4d0b8aa9bf9e.mov

(with sound - you can hear me clicking mouse and hitting keyboard buttons)

**What this PR does**

Replaces `.isMouseMode` with `.focus-visible` class which is a REVERSE of the former and is in line with WICG focus-visible polyfill.

Adjusts all styles that used `.isMouseMode` to use `.focus-visible` instead.

This makes it incredibly simple to ditch `useFocus` in favor of WICG focus-visible polyfill, with no additional changes to styles.

Adds `GELClickHandler` that does the reverse of GELKeyHandler: toggles back mouse mode (i.e. removes `.focus-visible` class). Consequently, ditches the automatic removal of `GELKeyHandler` listener after one invocation, since it needs to be invoked multiple times now

**Concerns:**

| Description | Mitigation |
| --- | --- |
| If page includes popular WICG focus-visible polyfill on its own, the scripts would be both handling `.focus-visible` class name. WICG's version could be less aggresive with adding `.focus-visible` class as it does additional heuristics. | `.gel-focus-visible` class name can be used for now until (and if) we switch to WICG focus-polyfill |

### DIFF
--- a/components/button-group/src/overrides/button.js
+++ b/components/button-group/src/overrides/button.js
@@ -74,7 +74,7 @@ const buttonStyles = () => {
 		},
 
 		// Focus state
-		'body:not(.isMouseMode) input:focus + &': {
+		'body.focus-visible input:focus + &': {
 			...PACKS.focus,
 		},
 	})[0];
--- a/components/core/blender/script.js
+++ b/components/core/blender/script.js
@@ -77,27 +77,31 @@ var GEL = (function GELCore() {
 
 				if (!hasScript) {
 					// Insert a script that:
-					// - adds the "isMouseMode" class to the body
+					// - removes the "focus-visible" class to the body if somehow present
 					// - listens for the tab key
-					// - when tab key is pressed removes the "isMouseMode" class and removes the listener
+					// - when tab key is pressed adds the "focus-visible" class and removes the listener
 					var scriptEl = document.createElement('script');
 					scriptEl.setAttribute('id', 'GELFocus');
 					scriptEl.text =
-						'function GELKeyHandler( event ) {' +
-						'	if( event.key === "Tab" ) {' +
-						'		document.getElementsByTagName("body")[ 0 ].classList.remove("isMouseMode");' +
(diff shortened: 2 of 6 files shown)


wojtekmaj:
@jonnystening Is there anything preventing you from merging this?

PR #810, opened 2021-09-15
Icon: add new cursor icons

jonnystening:
Adding four new icons

PR also fixes buildExports.js...
- Adding component package.json source keys and root package endpoints no longer required in preconstruct
- Ensure SVGs in subfolders (e.g. pictograms) can be processed, as well as those that dont use subfolders (e.g. icons and symbols)


jaortiz (approved):
LGTM

PR #847, opened 2021-11-22
a11y updates

jonnystening:
Just waiting to hear back from the others about text alignment in the situation below...
Screen Shot 2021-11-22 at 5 15 18 pm


jaortiz (approved):
LGTM

PR #850, opened 2021-11-25
Replace classnames with clsx 

### LINKED ISSUE #795: Consider clsx instead of classnames
clsx is a smaller, faster package for merging classnames, fully compatible with classnames API.

wojtekmaj:
classnames is not used directly in @westpac/progress-rope - we're using classNames from @westpac/core

Closes #795


jaortiz:
@wojtekmaj classnames is currently being used as part of another system to make the components backwards compatible with legacy systems at Westpac. I'm currently looking into reworking that entire system and will probably remove classnames altogether so I don't see the value in changing to clsx now. I will remove the double classnames dependency in another pr I am currently working on.

PR #889, opened 2022-09-06
Server render article pages instead

flexdinesh:
Current build process does not support fetching API data during the build since frontend and API builds are not sequential. This pull request switches the approach from statically building article pages to server rendering the pages where we fetch article home and article page data during server rendering.


jaortiz (approved):
LGTM

PR #891, opened 2022-09-12
Article title and card image changes

flexdinesh:
Two changes included in this PR —

1. Article card image should be used instead of page image in articles home and Admin UI editor
2. Article title should not be shown in place of article description in articles home


jaortiz (approved):
LGTM

PR #893, opened 2022-09-13
Increase nginx file upload size limit

flexdinesh:
Default nginx file upload limit is 1Mb. Some of the image files being uploaded for article pages are >1MB in size and we need to support that.

Changes:

- Increase nginx file upload size limit


dcousens (approved):
LGTM

jaortiz (approved):

PR #910, opened 2022-11-02
Export TS interfaces using existing prop types 

samithaf:
This is a draft version of how we can export TS interfaces using existing prop type definitions in the components. 

This approach will enable GEL consumers to use TS in their projects. In the long run we can simply convert GEL to TS but this will require more work.


samithaf:
IDE detect the TS interface definitions with the change.  

Screen Shot 2022-11-02 at 5 07 00 pm

PR #917, opened 2022-11-23
Updated components to export TypeScript interfaces using existing prop types.

PinkTaco97:


package.json
```diff
@@ -75,13 +75,13 @@
     "testRegex": "\\.spec.js$",
```
samithaf, line 75: Let's update the matcher to catch ts, tsx as well

components/__tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
```
samithaf, line 1: Do we need this config?

samithaf (approved):

PR #924, opened 2022-12-22
Initial tests for the Heading component

wove5:
Awaiting review by Samitha and Walter.


wove5:
Went back and applied  "format:fix"  and  "lint:fix" in new commit, then closed and reopened the PR after which auto checks passed.

components/heading/tests/unit/heading.spec.js
```diff
@@ -1,3 +1,43 @@
...
-	expect(true).toBe(true);
+import React from 'react';
+import { Heading } from '@westpac/heading';
+import { render } from '@testing-library/react';
+
+import { overridesTest } from '../../../../helpers/tests/overrides-test.js';
+import { nestingTest } from '../../../../helpers/tests/nesting-test.js';
+import { GEL } from '@westpac/core';
+import wbc from '@westpac/wbc';
+// import { Heading } from '@westpac/button-dropdown';
```
wpalombini, line 9: If this is not going to be used, please remove it, thanks.

components/heading/tests/unit/heading.spec.js
```diff
@@ -1,3 +1,43 @@
...
+
+// The default tests every component should run
+overridesTest({
+	name: 'heading', // the name has to be the package name without '@westpac/' scope
+	overrides: ['Heading'], // every single override root key
```
wpalombini, line 14: It seems that we are missing an override here: BrandHeading. Check the components/heading/src/overrides folder. You can also compare this with the Alert tests.

wpalombini (commented):
Hi David, thanks for working on adding more tests to it. I just let a couple of comments, otherwise, it looks good to me.

components/heading/tests/unit/heading.spec.js
wove5: Yes, shall do - look for next commit.  Line 9 came up in auto completion during coding, then I commented it out - forgot to remove it.

components/heading/tests/unit/heading.spec.js
wove5: overrides: ['BrandHeading', 'Heading'], // every single override root key
This is making the unit test fail. I need to discuss this with you further.

wove5: Ok, now the branch includes testing for BrandHeading component.

wpalombini (approved):
It looks good David!

PR #927, opened 2023-01-12
Updated the Brands & Core Utilities to export TypeScript Interfaces.

PinkTaco97:


components/core/src/useManagedState.ts
```diff
@@ -1,8 +1,12 @@
-import { ChangeEvent, useRef, useState } from 'react';
+import { useRef, useState } from 'react';
 
 import { devWarning } from './devWarning';
 
-export function useManagedState(controlledValue, defaultValue, onChange) {
+export function useManagedState<Tvalue>(
+	controlledValue: Tvalue,
+	defaultValue: Tvalue,
+	onChange: Function
```
kenjishiromajp, line 8: minor: I think onChange is expecting to be something like:
`onChange: (value: any, event: Event) => any;`

components/core/src/wrapHandlers.ts
```diff
@@ -1,4 +1,4 @@
-export const wrapHandlers = (consumerHandler, ourHandler) => (event) => {
+export const wrapHandlers = (consumerHandler: Function, ourHandler: Function) => (event: Event) => {
```
kenjishiromajp, line 1: minor(typescript): I think the handler could be type as:

```ts
type IHandler = (event: Event) => any;
export const wrapHandlers = (consumerHandler: IHandler, ourHandler: IHandler) => (event: Event) => {
}
```

kenjishiromajp (approved):

PR #930, opened 2023-01-17
feat(useInstanceID): refactoring to use useID instead

kenjishiromajp:
- refactoring to use useID instead of useInstanID
- Adding prettier in eslint since it doens't tell the error when linting


components/core/src/Core.tsx
```diff
@@ -5,13 +5,12 @@ import { Fragment } from 'react';
 import { useBrand } from './Brand';
 import { normalize as normalizeCSS } from './normalize';
 
-export interface ICoreProps {
+export interface CoreProps {
```
PinkTaco97, line 8: Shouldn't this interface be called "ICoreProps"?
```suggestion
export interface ICoreProps {
```

PinkTaco97 (approved):

PR #931, opened 2023-01-25
Feature/wlt 14195 unit tests autocomplete

kenjishiromajp:
- Unit test for autocomplete


wove5:
Good testing.  Should we test for the colored highlighting of the Select input outline when invalid prop is passed?

The isClearable prop? Is it necessary for rendering the clear button?

wove5:
On a second look, the isClearable prop has default value of true, hence its inclusion or exclusion makes not difference.

wove5:
I would like to know where the colouring for the Select input is coming from.  Colours look like they get set in the component overrides styles; examples are control, footer, menu and option modules. But cannot work out how COLORS is constructed `const { COLORS } = useBrand()` .  Also not able to manually effect a change in the colors to test how they can be modified.

wove5 (commented):
I'd like to know where you found the rgb(147, 144, 162) value, in:
`@@ -1,5 +1,11 @@
import DefaultPage from '../pages/default.page';

const defaultPage = new DefaultPage();
describe('Autocomplete', () => {
	before(() => {
		cy.visit(`http://localhost:8080/`);
		defaultPage.visit();
	});
	it('should contain a red border when there is a prop invalid', () => {
		defaultPage.invalidAutocomplete.get('[class$="Control"]').should('have.css', 'border', '1px solid rgb(147, 144, 162)');
	});
});`

I was looking for this but could not find.  Anyway, I like your tests that you've implemented with cypress.

PinkTaco97 (approved):

PR #934, opened 2023-01-27
Chore/header tests

wove5:
I have built a suite of tests for the Header component. There's a possibility I could add a test for scrolling behaviour later, but I'd like to submit this unit test for the Header now. Happy to receive feedback.


components/header/tests/unit/header.spec.js
```diff
@@ -1,3 +1,188 @@
...
+	});
+
+	test('should render its content', () => {
+		const text = 'Hello';
+		const content = <span data-testid="my-span">{text}</span>;
+		const SimpleHeader = () => (
+			<GEL brand={wbc}>
+				<Header>{content}</Header>
+			</GEL>
+		);
```
kenjishiromajp, line 43: minor(reafactor): I think this component could be outside and reused across the tests since it has been repeated.

kenjishiromajp (changes requested):

components/header/tests/unit/header.spec.js
wove5: Yes, ok, thanks for the tip; made that change. Please take a squiz.

PR #944, opened 2023-02-16
Chore/button tests

wove5:
- Please note my commented out line 98.  Perhaps you might know why it would not work for me, whereas getByRole('img') had worked before in my previous test branch.

- I was going to try and build a test for the Content component, but I could not work out how to import that component into a spec file.

- Interestingly, the @westpac/button/index.ts exports Button and ButtonContext components, but not the Content component.  ButtonContext is not defined in its own file but is defined inside the Button component file.

- There also appears to be another component named Text.tsx inside the button repo which I have not  tried testing; should I test it? 
- Like the Content component, the Text component is not exported out of @westpac/button/index.ts.

- I think I should test Content because it has one override inside it, namely Icon, in addition to the usual override of the component that is the subject of the test.
- So I am unsure whether I should write tests for the Content, Text and ButtonContext components; and how I should go about writing them if I need to.


components/button/tests/unit/button.spec.js
```diff
@@ -0,0 +1,105 @@
...
+		const theButton = container.querySelector('button[disabled]');
+		expect(theButton).toBeInTheDocument();
+		expect(theButton).toHaveTextContent(someText);
+	});
+
+	test('should show normal, default, button color when not active', () => {
+		const { getByTestId } = render(<SimpleButton data-testid="my-button" />);
+		const theButton = getByTestId('my-button');
+		// Property 'look' has default value 'hero', so check for the hero color
+		expect(theButton).toHaveStyle('background-color: rgb(31, 28, 79)');
```
samithaf, line 54: Colour theme can change. Please don't use colours directly in unit tests. We should let the UX to change the colours when it's needed.

components/button/tests/unit/button.spec.js
```diff
@@ -0,0 +1,105 @@
...
+
+const SimpleButton = (props) => (
+	<GEL brand={wbc}>
+		<Button {...props}>{props.children}</Button>
+	</GEL>
+);
+
+// Component specific tests
+describe('Button component', () => {
+	const originalLog = console.log;
```
samithaf, line 31: Instead of overriding the console.log, create a simple React component which uses the button component. Then within the component you could change text or equal to see if the on click is working.

samithaf (changes requested):

components/button/tests/unit/button.spec.js
wove5: Please see my changes in the following commit: 02544070
I left the test that makes a console.log and then spies on it - I don't think this one actually overrides console.log, like the lines I removed did. But I'll remove the test if you want it gone.

samithaf (approved):

PR #946, opened 2023-02-22
Feature: generate proptypes from typescript

kenjishiromajp:
- Using interface in order to generate the the proptypes using the `yarn prop-types` command.


components/a11y/demos/dev/10-visuallyHidden.js
```diff
@@ -1,5 +1,3 @@
-/** @jsx jsx */
```
samithaf, line 1: I wonder it would be easier if we replace `/** @jsx jsx */` with `/** @jsxImportSource @emotion/react */`

samithaf (changes requested):

components/a11y/demos/dev/10-visuallyHidden.js
samithaf: This is fine. As per the discussion it's better to manage it via Babel config since it's more robust.

samithaf (approved):

PR #949, opened 2023-03-01
Changed incorrect component Button to ButtonDropdown in nesting test

wove5:
Also removed Panel from overrides test which was causing a warning.


wove5:
@kenjishiromajp , there is a warning about refs for the Panel component when it's included in the overrides test. The warning is:
### Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
Can you take a look at this with me when you have time.

samithaf (approved):

PR #956, opened 2023-03-14
Chore/wlt 20989/button group tests

wove5:
One extra thing to note but not mentioned in the commit msg's, is that tsconfig.json had a small change to make the  cypress tests work - see the file diff.


samithaf:
@wove5 Can you please check if we have enabled Prettier and ESLint for these files?

wove5:
@samithaf , are these lines in root package.json what you mean:
https://github.com/WestpacGEL/GEL/blob/1b8cc193a8b582ba4251dfea294f8c1f10ab6917/package.json#L50-L56

image

eslint and prettier are running manually.
and eslint automatically on push to GH

image

When I tried pushing some work with broken code in button-group.cypress.js two days ago, the auto eslint blocked the push.

wove5:
@kenjishiromajp , Vinod and myself have noticed a consistent warning about a failed prop type in the browser console whenever the Button-Group is rendered, both in dev mode and when running tests (jest & cypress). Can you take a look at this when you get a chance. The warning is titled:
### react_devtools_backend.js:2655 Warning: Failed prop type: The prop `inputProps` is marked as required in `ButtonGroupItem`, but its value is `undefined`.
It would be good if Vinod and myself can go through it together with you.

PR #961, opened 2023-04-03
chore: create unit & cypress tests for 3 components in text-input pac…

wove5:
…kage; modify global nesting-test

Cypress tests created for TextInput and Textarea components. Jest/Testing-Library tests created for TextInput and TextInputWithButton components.
The general GEL-repo-wide nesting-test.js unit test has had the insertion of a string literal child into the component undergoing testing removed from the test. The test previously tested for the existence of the string literal as a text node which would fail for elements that did not accept children such as button and input.

### DIFF
--- a/components/text-input/tests/unit/text-input-with-button.spec.js
+++ b/components/text-input/tests/unit/text-input-with-button.spec.js
@@ -1,0 +1,44 @@
+import { GEL } from '@westpac/core';
+import { render } from '@testing-library/react';
+import { screen } from '@testing-library/dom';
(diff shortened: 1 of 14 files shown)


wove5:
Tom said he'd take a look at this tomorrow, Wed 5th

components/text-input/tests/unit/text-input-with-button.spec.js
tomjackking, line 1: I would test that the button works when passing the btnOnClick prop for this component

PR #972, opened 2023-04-14
Label Unit Tests

tomjackking:
- added TS tests for label
- added new tests for previous components I worked on to increase coverage
- minor change to nesting-test and overrides-test to export to avoid TS issues
- minor changes to WBC types to add key to colors to avoid TS issues in test

### DIFF
--- a/brands/WBC/src/types.d.ts
+++ b/brands/WBC/src/types.d.ts
@@ -4,7 +4,6 @@ declare module '@westpac/wbc' {
 		code: string;
 	};
 	type Colors = {
-		[key: string]: string;
 		tints: {
 			background5: string;
 			background10: string;
--- a/brands/WBC/src/types.d.ts
+++ b/brands/WBC/src/types.d.ts
@@ -4,6 +4,7 @@ declare module '@westpac/wbc' {
 		code: string;
 	};
 	type Colors = {
+		[key: string]: string;
 		tints: {
 			background5: string;
 			background10: string;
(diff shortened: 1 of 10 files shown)


brands/WBC/src/types.d.ts
```diff
@@ -4,6 +4,7 @@ declare module '@westpac/wbc' {
 		code: string;
 	};
 	type Colors = {
+		[key: string]: string;
```
samithaf, line 7: Colours needs to pre-defined set of key-pair values. So please remove this

tomjackking: This line has been removed

components/form/tests/unit/fieldset.spec.js
```diff
@@ -41,6 +41,14 @@ describe('Fieldset component', () => {
 		expect(getByText(/test hint/).getAttribute('class')).toMatch(/form-hint/gi);
 	});
 
+	test('should render Fieldset component when hint prop contains function', () => {
+		const handleClick = jest.fn(() => {});
+		const { getByTestId } = render(
+			<SimpleFieldset hint={handleClick} data-testid="test-fieldset" />
```
wove5, line 47: Where does data-testid get used?

components/label/tests/unit/label.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+nestingTest({
+	name: 'label',
+	Component: (props: any) => <Label look="primary" value="default" {...props}></Label>,
+});
+
+function SimpleLabel(props: any) {
+	return (
+		<GEL brand={wbc}>
+			<Label look={props.look} {...props} data-testid="test-label" />
```
wove5, line 23: Where does data-testid get used?

samithaf (approved):

PR #973, opened 2023-04-18
Chore/wlt 24622/fork tests

wove5:


helpers/tests/overrides-test.js
```diff
@@ -26,7 +26,7 @@ function overridesTest({ name, overrides, Component }) {
 				const withOverrides = cloneDeep(wbc);
 				tokenOverrides = {
 					[override]: {
-						styles: (styles) => ({ ...styles, content: styleText }),
+						styles: (styles) => ({ ...styles, content: styleText, display: 'block' }),
```
soobin93, line 29: This might could cause an issue in the future where the override component shouldn't be `display: 'block` by default as it's sort of forcing in the global file.
We might should find another way to handle this issue from the specific unit test or we should create another ticket upgrading `overrides-test` file.

What do you guys think?

tomjackking (approved):

PR #976, opened 2023-04-21
Chore/WLT-24625: Unit tests for FormPod components

soobin93:


tomjackking (approved):

components/form-pod/tests/unit/form-pod-panel-footer.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+overridesTest({
+	name: 'form-pod',
+	overrides: ['Footer'],
+	Component: (props: any) => {
+		const LeftItem = <div>Left Item</div>;
+		const RightItem = <div>Right Item</div>;
+		return <PanelFooter left={LeftItem} right={RightItem} {...props} />;
+	},
+});
```
wove5, line 18: `overrides: ['Footer'],`
I was wondering if this could or should include 'FooterItem'

wove5: When adding 'Footer', it failed on the FooterItem overrides, but when rendering just one FooterItem, by removing either left or right prop, the test passed

soobin93: hi @wove5 , currently I could not find an alternative way to test `FooterItem`, without updating the `overridesTest` function due to the reason it's currently giving a same `testid` to both of those elements.
Can we skip `FooterItem` for now and add it to the TODO list once `overridesTest` function gets improved?

soobin93: just left a TODO comment on it :)
will handle this matter in different PR shortly

components/form-pod/tests/unit/form-pod.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+	it('renders children, preheading text and heading text', () => {
+		render(<SimpleFormPod {...defaultProps} />);
+		expect(screen.getByText('FormPod Content')).toBeInTheDocument();
+		expect(screen.getByText('FormPod Preheading')).toBeInTheDocument();
+		expect(screen.getByText('FormPod Heading')).toBeInTheDocument();
+	});
+
+	it('renders header icon when passed', () => {
+		render(<SimpleFormPod {...defaultProps} />);
+		expect(defaultProps.icon).toHaveBeenCalled();
```
wove5, line 54: I like that - very cool; the icon has been made a "dummy" jest function and checking that it is called is equivalent to being rendered. I'll approve your PR now.

wove5 (approved):

components/form-pod/tests/unit/form-pod.spec.tsx
soobin93: Thanks :)

PR #979, opened 2023-04-27
Icon unit tests

tomjackking:
- added test that covers every icon
- made minor change to package and converted index.js to ts to remove TS warn/issue that would pop up when importing icons
- NOTE: the above change will require you to run yarn again when pulling develop


components/icon/tests/unit/icons.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+	overrides: ['Icon'],
+	Component: (props: any) => <Icon copyrightYear="2020" size="medium" {...props} />,
+});
+
+nestingTest({
+	name: 'icon',
+	Component: (props: any) => <Icon {...props} />,
+});
+
+function SimpleIcon(props: any) {
```
samithaf, line 21: Would it be possible to define the component contract instead of using "any"?

soobin93 (approved):

components/icon/tests/unit/icons.spec.tsx
tomjackking: Just leaving a comment here after discussing through teams with @samithaf and @soobin93. The Icon component is not a TS component and is still JS so there is no interface to import and can be used would be better to convert Icon to TS but that should be a separate task. I have updated the Label tests using the interface exported from that component and we will be going forward doing that for future TS tests now we have a working example.

wove5: Well explained. The PR is looking good.

PR #981, opened 2023-05-02
Pictogram Unit Tests

tomjackking:
- added tests for the variations Pictogram components
- minor tests for base Pictogram component


components/pictogram/tests/unit/pictograms.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+overridesTest({
+	name: 'pictogram',
+	overrides: ['Pictogram'],
+	Component: (props: any) => <Pictogram copyrightYear="2023" {...props} />,
+});
+
+nestingTest({
+	name: 'pictogram',
+	Component: (props: any) => <Pictogram {...props} />,
+});
```
wove5, line 18: Just curious to know how the Pictogram renders without throwing a warning about missing the copyrightYear prop.

components/pictogram/tests/unit/pictograms.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+		let typedVal: Function;
+		if (typeof val === 'function') {
+			typedVal = val;
+		}
+		test(`should render ${key} with testing prop options`, () => {
+			const TestPictogram = (props: PictogramProps) => {
+				return typedVal({ ...props });
+			};
+			const DefaultTestPictogram = () => {
+				return typedVal({} as PictogramProps);
```
wove5, line 45: As per above comment, regarding the missing copyrightYear prop, but in DefaultTestPictogram this time

components/pictogram/tests/unit/pictograms.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+				</GEL>
+			);
+			expect(container).toBeInTheDocument();
+		});
+	}
+
+	test('should render Pictogram with correct height and width when width passed in', () => {
+		const { getByTestId } = render(
+			<SimplePictogram
+				width={[100, 150]}
```
wove5, line 62: Hi Tom, I was wondering if it is possible to test for the 150 width size? Does the array of values correspond to media query size responsiveness? I suppose mq responsiveness might not be possible to test with jest and the virtual DOM. I don't understand why in your spec file it chooses the 100 as default on render and not 150.

wove5 (approved):

components/pictogram/tests/unit/pictograms.spec.tsx
tomjackking: Discussed on Teams

components/pictogram/tests/unit/pictograms.spec.tsx
tomjackking: Discussed on Teams

components/pictogram/tests/unit/pictograms.spec.tsx
tomjackking: Every value represents a different break point, I haven't found a way to test other breakpoints on jest but I think that is more of an integration testing thing to do. I assume just tests what ever the first breakpoint is, as often times with some code I've tested that involves using breakpoints it will have null as the first option and null is the value it will use.

components/pictogram/tests/unit/pictograms.spec.tsx
tomjackking: From Teams: Spreading the props in the nesting test will make it use the default value for required props if they exist.

components/pictogram/tests/unit/pictograms.spec.tsx
tomjackking: From Team: This is forcing the component to use default props to test the default props as there were issues where it was expecting the required props even with spreading props so I couldn't test defaults unless I did what I did here.

helpers/tests/overrides-test.js
tomjackking, line 1: Explanation of this update:
For some components duplicate props were showing up outside of the state props causing them to be spread onto the childWrapper div. This would cause warnings to show up in the log. After some digging I found that only data-attribute, children and className were the only props that were required to be on the div so destructured those instead and removed state and ...rest destructuring as state no longer needed to exist for the sole purpose of removing state props from the spread.

soobin93 (approved):

PR #982, opened 2023-05-04
Symbol Unit Tests

tomjackking:
- added unit tests for symbol components


components/symbol/src/overrides/symbol.js
```diff
@@ -57,7 +58,7 @@ const symbolStyles = (_, { symbol, width, height, viewBoxWidth, viewBoxHeight })
 
 const blenderStyles = (_, { symbol, width, height, viewBoxWidth, viewBoxHeight }) => {
 	const props = { symbol, width, height, viewBoxWidth, viewBoxHeight };
-	const baseStyles = symbolStyles(_);
+	const baseStyles = symbolStyles(_, defaultProps);
```
tomjackking, line 61: blenderStyles was not working, icon blenderStyles is using a similar implementation and defaultProps was passed in for this same line for icon, did the same here and it started working.

components/symbol/tests/unit/symbol.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+	Component: (props: any) => <Symbol copyrightYear="2020" align="center" {...props} />,
+});
+
+nestingTest({
+	name: 'symbol',
+	Component: (props: any) => <Symbol {...props} />,
+});
+
+// using any typing on props as symbol is not TS and does not have an interface
+function SimpleSymbol(props: any) {
```
tomjackking, line 22: Just so it's extra clear, like icon, symbol is not a TS file so it has no type interface to use which is why any is used here. Could be updated in the future if the symbol component gets converted to TS.

soobin93 (approved):

components/symbol/tests/unit/symbol.spec.tsx
```diff
@@ -0,0 +1,83 @@
...
+import wbc from '@westpac/wbc';
+import { overridesTest } from '../../../../helpers/tests/overrides-test.js';
+import { nestingTest } from '../../../../helpers/tests/nesting-test.js';
+import * as symbols from '@westpac/symbol';
+import { Symbol } from '../../src/Symbol.js';
+import { blenderSymbol } from '../../src/overrides/symbol.js';
+
+overridesTest({
+	name: 'symbol',
+	overrides: ['Symbol'],
```
wove5, line 12: Should we include the Svg component in the overrides test?

wove5: Actually, it through a warning when I added Svg. Is it better to leave out then?
  console.error
    Warning: The tag <metadata> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter.

samithaf (approved):

PR #988, opened 2023-05-10
Chore/wlt 26341/switch tests

wove5:
The onChange event-handler was not able to be triggered in the unit tests for the Switch component.  Whilst other elements like button can have their event handlers triggered in testing, toggle switches do not seem to expose their event-handlers to the jest testing framework, therefore other means were used to test the responsiveness of the Switch such monitoring the value of the checked property and the background color of the toggle.


components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+		const { container } = render(<SimpleSwitch label="A Switch" />);
+		expect(container).toBeInTheDocument();
+	});
+
+	test('should be able to have its checked valued changed', () => {
+		render(<SimpleSwitch label="A Switch" />);
+		const switchInput = screen.getByRole('checkbox');
+		expect(switchInput).toHaveProperty('checked');
+		expect(switchInput).toHaveProperty('checked', false);
+		fireEvent.change(switchInput, { target: { checked: true } });
```
soobin93, line 39: could we use `userEvent` from `@testing-library/user-event` instead of fireEvent? as it's more close to actual user's action.
Further explanation here: [Link](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event)

and make sure that you need to use async and await for userEvent

soobin93 (changes requested):

components/switch/tests/unit/switch.spec.tsx
wove5: Yes, can do that.  I'll make that change.  Thanks for your input.

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+		expect(switchInput).toHaveProperty('checked', true);
+		fireEvent.change(switchInput, { target: { checked: false } });
+		expect(switchInput).toHaveProperty('checked', false);
```
tomjackking, line 42:
```suggestion
		expect(switchInput).not.toBeChecked();
		fireEvent.change(switchInput, { target: { checked: true } }); //Soobins suggestion
		expect(switchInput).toBeChecked();
		fireEvent.change(switchInput, { target: { checked: false } }); //Soobins suggestion
		expect(switchInput).not.toBeChecked();
```

components/switch/tests/unit/switch.spec.tsx
tomjackking: On top of Soobin's suggestion, you may only need to do fireEvent.click() which I think is the simplest solution if you didn't already try click()

wove5: userEvent & fireEvent have both been used in the test, just to add some variety, and also to see that both work and have the same effect.  The necessity of async, await, waitFor in the expectations with userEvent makes a bit more work for the userEvent click, so agree with you both: fireEvent simplifies things greatly and just works, while userEvent is very thorough and can test for a lot more things.  I guess they both have their place.

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+		);
+		const switchText = getByLabelText('A Switch');
+		const switchInput = getByRole('checkbox');
+		const switchToggle = container.querySelector('span[class$=-switch-toggle]') as HTMLElement;
+		const switchParent = getByTestId('switch-parent');
+		const user = userEvent.setup();
+		user.click(switchText);
+		await waitFor(() => {
+			expect(switchInput).toBeChecked();
+			expect(switchInput).toHaveProperty('checked', true);
```
tomjackking, line 72:
```suggestion
```
Doesn't need to exist with toBeChecked

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,134 @@
...
+			expect(handleChange).toHaveBeenCalledTimes(1);
+		});
+		user.click(switchInput);
+		await waitFor(() => {
+			expect(switchInput).not.toBeChecked();
+			expect(switchInput).toHaveProperty('checked', false);
```
tomjackking, line 78:
```suggestion
```

Doesn't need to exist with not.toBeChecked

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		await user.click(switchText);
+		expect(window.getComputedStyle(switchToggle).getPropertyValue('background-color')).toEqual(
+			checkedSwitchTogglebackgroundColor
+		);
+		await user.click(switchToggle);
+		expect(window.getComputedStyle(switchToggle).getPropertyValue('background-color')).toEqual(
+			uncheckedSwitchTogglebackgroundColor
+		);
+		expect(handleChange).toHaveBeenCalledTimes(4);
+	});
```
tomjackking, line 99: You don't need to test every component that can be clicked. I would just be testing switchInput

components/switch/src/Switch.tsx
```diff
@@ -120,7 +120,11 @@ export const Switch = forwardRef<HTMLInputElement, SwitchProps>(
 			setChecked(isChecked);
 		}, [isChecked]);
 
-		const handleChange = () => wrapHandlers(onChange, () => setChecked(!checked));
+		const handleChange = (event: any) =>
```
tomjackking, line 123: Because of these changes you will need to do a changeset file

components/switch/tests/unit/switch.spec.tsx
wove5: Ok Tom.  I pushed just before you made both this and that previous comment which takes into account Soobin's suggestion.  I will make the changes now suggested in both your comments

components/switch/src/Switch.tsx
wove5: Yes, the changeset is what I wanted to talk with you about, when I pinged you late Fri.  I will try and chase you up in the afternoon

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		const handleChange = jest.fn();
+		const { container, getByRole, getByTestId, getByLabelText } = render(
+			<SimpleSwitch label="A Switch" onChange={handleChange} />
+		);
+		const switchText = getByLabelText('A Switch');
+		const switchInput = getByRole('checkbox');
+		const switchToggle = container.querySelector('span[class$=-switch-toggle]') as HTMLElement;
+		const switchParent = getByTestId('switch-parent');
+		const user = userEvent.setup();
+		expect(switchInput).toHaveProperty('checked');
```
tomjackking, line 44:
```suggestion
```
not needed if you have not.toBeChecked()

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		expect(switchInput).not.toBeChecked();
+		expect(switchInput).toHaveProperty('checked', false);
+		await user.click(switchText);
+		expect(switchInput).toBeChecked();
+		expect(switchInput).toHaveProperty('checked', true);
```
tomjackking, line 49:
```suggestion
```
not needed if you have toBeChecked()

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		expect(handleChange).toHaveBeenCalledTimes(1);
+		await user.click(switchInput);
+		expect(switchInput).not.toBeChecked();
+		expect(switchInput).toHaveProperty('checked', false);
```
tomjackking, line 53:
```suggestion
```
as above comments

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		expect(handleChange).toHaveBeenCalledTimes(2);
+		await user.click(switchToggle);
+		expect(switchInput).toBeChecked();
+		expect(switchInput).toHaveProperty('checked', true);
```
tomjackking, line 57:
```suggestion
```
as above comments

components/switch/tests/unit/switch.spec.tsx
```diff
@@ -0,0 +1,100 @@
...
+		expect(handleChange).toHaveBeenCalledTimes(3);
+		await user.click(switchParent);
+		expect(switchInput).not.toBeChecked();
+		expect(switchInput).toHaveProperty('checked', false);
```
tomjackking, line 61:
```suggestion
```
as above comments

wove5: Thanks Tom, I have removed those in my latest commit pushed now. There's one thing that I might revert in the commit; that is a change to proptypes in Compactor.tsx and Repeater.tsx, due to running "Compactor.tsx and Repeater.tsx". Soobin knows and will take a look after lunch.

wove5:
Added in a test for blender overrides to increase test coverage.

Duwon98 (approved):

PR #999, opened 2023-05-17
List unit tests

Duwon98:


soobin93 (approved):

components/list/tests/unit/list.spec.tsx
```diff
@@ -0,0 +1,168 @@
...
+	});
+
+	it('renders nested List correctly', () => {
+		const WBCNestedList = () => (
+			<GEL brand={wbc}>
+				<List
+					data={[
+						'item1',
+						{
+							text: 'item2',
```
wove5, line 91: I just wonder where the "text" prop is in the Item?

Duwon98: It is from data props. It would be easy to understand if you refer to data-driven files from the example directory.

PR #1008, opened 2023-05-25
GEL updates Icon

Duwon98:
Update base Icon component to export Typescript interface

Update all icons to export TS interfaces.

Migrate all Icons files from js to tsx


components/icon/src/AccessibilityIcon.tsx
```diff
@@ -0,0 +1,67 @@
+import PropTypes from 'prop-types';
+import { Icon, IconProps } from './Icon';
+
+export const AccessibilityIcon = ({
+	assistiveText = 'Accessibility',
+	copyrightYear = '2020',
+	size = 'medium',
+	overrides,
+	...props
+}: Omit<IconProps, 'icon'>) => (
```
wove5, line 10: Where is the 'icon' type encountered? If it is encountered somewhere, is that the reason for using the Omit constructor?

wove5 (changes requested):

components/icon/src/AccessibilityIcon.tsx
Duwon98: It means we are using IconProps type but not using icon from IconProps. 
All the icons are from baseIcon. as you can see below, baseIcon component is using icon prop. 
Omit has been used to prevent duplication

wove5 (approved):

components/icon/src/AccessibilityIcon.tsx
```diff
@@ -0,0 +1,68 @@
...
+	// |     To update them edit TypeScript types and run "yarn prop-types"  |
+	// ----------------------------------------------------------------------
+	/**
+	 * String to use as the `aria-label` for the icon. Set to an empty string if you
+	 * are rendering the icon with visible text to prevent accessibility label
+	 * duplication.
+	 *
+	 * Defaults to the icon name e.g. `BusinessPersonIcon` --> "Business Person"
+	 */
+	assistiveText: PropTypes.string,
```
kenjishiromajp, line 39: @Duwon98 the proptype color is missing.

kenjishiromajp (changes requested):

components/icon/src/AccessibilityIcon.tsx
Duwon98: Hey Kenji,
Changes have been made. Could you have a look? @kenjishiromajp @soobin93

kenjishiromajp (approved):

PR #1010, opened 2023-05-26
Refactoring Modal

tomjackking:
- refactored/tested modal to work with movemoney without warnings/hydration errors
- removed unused refs that were causing warnings
- fixed escape key not closing modal


components/modal/src/Modal.tsx
```diff
@@ -167,9 +160,6 @@ export const Modal = ({
 		Backdrop: { component: Backdrop, styles: backdropStyles, attributes: backdropAttributes },
 	} = overrideReconciler(defaultOverrides, tokenOverrides, brandOverrides, componentOverrides);
 
-	const modalRef = useRef(null);
-	const headingRef = useRef(null);
```
tomjackking, line 171: Theses refs weren't being used anywhere and were causing warning to pop up in movemoney. Based on git history it seems they were to be used for focusing inside the modal but there is a library that seems to be doing that now.

components/modal/src/Modal.tsx
```diff
@@ -182,84 +172,81 @@ export const Modal = ({
...
 	}, [onClose]);
 
-	// on escape close modal
-	const keyHandler = useCallback(
-		(e: KeyboardEvent) => {
-			if (dismissible && open && e.keyCode === 27) handleClose();
-		},
-		[dismissible, handleClose, open]
-	);
-
```
tomjackking, line 192: This code was not working, it was meant to close the modal if it was dismissable and someone hit escape. The react-focus-on library FocusOn component has a prop that enables this functionality, that has been implemented and tested, it works

components/modal/src/Modal.tsx
```diff
@@ -182,84 +172,81 @@ export const Modal = ({
...
-	// bind key events
 	useEffect(() => {
-		window.document.addEventListener('keydown', keyHandler);
-		return () => {
-			window.document.removeEventListener('keydown', keyHandler);
-		};
```
tomjackking, line 198: No longer needed for same reason as removing keyHandler function

components/modal/src/Modal.tsx
```diff
@@ -182,84 +172,81 @@ export const Modal = ({
...
-						>
-							<ModalDialog
-								state={state}
-								{...modalDialogAttributes(state)}
-								css={modalDialogStyles(state)}
+	// These styles call hooks and will create hook call warning if called directly in component
+	const modalDialogStyle = modalDialogStyles(state);
+	const headerStyle = headerStyles(state);
+	const headingStyle = headingStyles(state);
+	const closeBtnStyle = closeBtnStyles(state);
```
tomjackking, line 183: These need to be called outside of the component as they contain the useBrand hook, if they are called like some of the others inside the component it will create warnings about unexpected hook calls

components/modal/src/Modal.tsx
```diff
@@ -182,84 +172,81 @@ export const Modal = ({
...
+
+	/**
+	 * Tab selection causes re-render of modal if portal is separated out as component and passed in
+	 * however if done as seperate component won't have the hook call warnings
+	 **/
+	return (
+		<Fragment>
+			{mounted &&
+				ReactDOM.createPortal(
```
tomjackking, line 192: I decided on this solution as it was easier to solve the issue of the unexpected hook calls rather than the tab selection causing a weird interaction with the modal causing it to make a new Modal/re-rendering then having to press tab again to select. Once second modal was closed it would open the first one again.

components/modal/src/Modal.tsx
```diff
@@ -182,84 +172,81 @@ export const Modal = ({
...
+					<GEL brand={brand}>
+						<ModalContext.Provider value={{ state }}>
+							<FocusOn
+								enabled={open}
+								onEscapeKey={() => {
+									if (dismissible) handleClose();
+								}}
```
tomjackking, line 199: this prop replaces the code above that was meant to do the same thing

components/modal/src/Modal.tsx
benweier: This should be the other way around. A hook must be called within a component otherwise React will complain about unexpected calls.

tomjackking: I may have worded that the wrong way, if they are called inside the return it was causing unexpected calls warnings. They are still being called in the component, here is an example of what was happening:
image
image

.changeset/clean-tomatoes-march.md
```diff
@@ -0,0 +1,5 @@
+---
+'@westpac/modal': minor
+---
+
+refactored modal to work with movemoney, also made minor changes to fix escape exiting modal and removing unused refs
```
samithaf, line 5: @tomjackking please remove the word "movemoney"

PR #1024, opened 2023-06-05
Fix/website work

jaortiz:


samithaf:
@jaortiz Can we update the Pre-requisites mentioned here too? https://gel.westpacgroup.com.au/design-system/development/get-started?b=WBC ( to React 18.2 + )

PR #1026, opened 2023-06-06
feat(WLT-28582): product tiles with arrow option

kenjishiromajp:
> Another example Kate has addressed:
- Added the option to have an arrow
image


jaortiz (approved):
LGTM

PR #1027, opened 2023-06-07
Feature/input field review

jaortiz:


components/input-field/src/components/InputField/InputField.tsx
```diff
@@ -16,12 +16,28 @@ export const useInputFieldContext = () => {
 	return context;
 };
 
-const InputFieldWrapper = ({ children, ...props }: { children: ReactNode }) => (
-	<div css={{ marginBottom: '1.875rem' }} {...props}>
+const Fieldset = ({ children, ...props }: { children: ReactNode }) => (
+	<fieldset role="group" css={{ border: 'none', margin: 0, padding: 0 }} {...props}>
```
ChrisLaneAU, line 20: One thing we should add is a `<legend>` element, populated by the `label` prop of the `InputField`. That way, when we're a fieldset, the label that is rendered becomes the group's label. We would then need to add an `aria-label` to the `<Input />`.

As an example, if you had a field for entering the amount someone wants to send overseas, you might have the following parts:
1. Group label: visible label of `Amount to send`
2. Text input for the amount: hidden label of `Amount`
3. Select input for the currency: hidden label of `Currency`

We could use our component like this:
```
<InputField isFieldset label="Amount to send">
	<Input aria-label="Amount" />
	<InputAfter>
		<Select aria-label="Currency" size="medium" inline={false} invalid={isInvalid}>
			<option>USD</option>
			<option>EUR</option>
			<option>JPY</option>
		</Select>
	</InputAfter>
</InputField>
```

And we'd loosely end up rendering markup something like this:
```
<fieldset>
	<legend>Amount to send</legend>
	<input aria-label="Amount" />
	<select aria-label="Currency">
		<option>USD</option>
		<option>EUR</option>
		<option>JPY</option>
	</select>
</fieldset>
```

This way, when a SR user enters the fieldset they'd get the label read out as the group's label and each input would have its own more specific label.

components/input-field/examples/50-scenarios.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+	const addOnAfterMap = {
+		idle: (
+			<InputAfter>
+				<Button look="hero" onClick={onClickCheckBtn}>
+					Check
+				</Button>
+			</InputAfter>
+		),
+		loading: <InputAfter icon={RefreshIcon} />,
```
ChrisLaneAU, line 78: Ahh sorry about this - it wasn't as polished as I thought it was, e.g. this should probably rotate and announce that it's loading but I think, for demo purposes of state changes we've conveyed the message to any consumers that want to pick it up. I'm happy to leave this as-is, if you are?

ChrisLaneAU (commented):
Looking great, just some small things. The <legend> stuff is probably the most important.

components/input-field/examples/50-scenarios.tsx
jaortiz: Yea happy to leave as is

ChrisLaneAU (approved):
Love it, great work. Thanks for all this!

PR #1031, opened 2023-06-14
FlexiCell test updates

ChrisLaneAU:
Updating the tests based on the recent changes by @kenjishiromajp. Note: these are only functional API tests.


components/circle/tests/unit/circle.spec.tsx
```diff
@@ -1,30 +1,34 @@
...
 				<GEL brand={wbc}>
-					<Circle>child</Circle>
+					<Circle>
+						<div data-testid="mock-child">child</div>
+					</Circle>
 				</GEL>
 			);
 
-			expect(getByText('child')).toBeVisible();
+			expect(screen.getByTestId('mock-child')).toBeVisible();
```
samithaf, line 17: @ChrisLaneAU getByTestId is discouraged and might need to use as the last resort, https://testing-library.com/docs/queries/bytestid

ChrisLaneAU: Hi Samitha! I agree that `getByTestId` is a last resort for selecting components because it does not reflect how users interact with the UI. In this particular style of test, where I'm asserting that any child is rendered, I believe `getByTestId` is a good fit because:
1. We want to know that we're specifically selecting the child being passed in - `getByText` could pick up the wrong text string, e.g. one that exists in the component (unlikely here but a good general approach).
2. The type of child doesn't matter, a `div` is generic and succinct.
3. We haven't applied the `data-testid` to a component that we're shipping so there's no worry about conflict in consumer codebases, it's only applied to the `div` created in this test file.
4. We could use another DS component or even something like `<button>mock button</button>` and select with `getByRole('button', { name: 'mock button' })` but I don't think that these have any more value than the `getByTestId` approach.

I'm happy to change this test if you think there's a better approach that offers more value.

samithaf: Sure, that's fair enough :)

samithaf (approved):

PR #1037, opened 2023-06-25
Feature/wlt 28582 patch

samithaf:
- Use CSS.Properties['fontWeight'] from csstype package instead of React. Fixes prop type warning
- Removed nested DIV tag within P tag. Fixes Cypress E2E warning


ChrisLaneAU (approved):
Thanks Samitha :)

